PR #38, opened 2017-04-04
[HOT-FIX] Temporary remove session cart subscriber

lchrusciel:
This fix is needed to make request stateless. This is just a temporary solution because it would break Sylius Shop.


pjedrzejewski:
👍

PR #44, opened 2017-04-04
[Config] Remove config file

lchrusciel:
This file is useless because `fos_rest.format_listener.rules` has to be defined only once. It is already covered in README file


pjedrzejewski:
👍

PR #45, opened 2017-04-04
[RFC] Rename CartViewTest to CartShopAPITest

lchrusciel:
I would be more understandable if we rename this class


pjedrzejewski:
👍

PR #57, opened 2017-04-19
[User] Register a new customer

### LINKED ISSUE #7: As a Customer, I want to register
* Create an account in the store

```
POST /register
```

lchrusciel:
Closes #7 

This endpoint returns an object, because of usage of ResourceController. But later, when we will move to Command Bus pattern it will be refactored anyway. 

Based on #60


pjedrzejewski:
Thanks Łukasz!

PR #77, opened 2017-05-05
[Product] [Cart] Change price contract

### LINKED ISSUE #68: Price handler
Currently, we are returning only one price on the product. Because of original pricing and in order to not break a contract in future we should go with and object instead. 
This:
```json
    "price": 1999
```
should be changed to:
```json
    "price": {
        "current": 1999
    }
```

lchrusciel:
All prices are unified to new model:
```json
{
    "price": {
        "current": 12313
    }
}
```

Resolves #68


tests/Responses/Expected/product/simple_product_details_page.json
```diff
@@ -9,7 +9,9 @@
             "name": "Logan Mug",
             "axis": [],
             "name_axis": [],
-            "price": 1999,
+            "price": {
+				"current": 1999
```
pjedrzejewski: Indentation seems to be messed up here in other json files. :(

PR #80, opened 2017-05-05
[Taxonomy] [Product] Add taxon code to product response

lchrusciel:
Based on #77 

Resolves #70

PR #81, opened 2017-05-05
[Taxon] Refactor Taxon controller and extract factory

lchrusciel:
During work on #80, I've decided to extract some of the common logic.

### DIFF
--- a/src/Factory/TaxonViewFactory.php
+++ b/src/Factory/TaxonViewFactory.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace Sylius\ShopApiPlugin\Factory;
+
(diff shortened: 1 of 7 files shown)


src/Factory/TaxonViewFactory.php
```diff
@@ -0,0 +1,43 @@
...
+     * @var ImageViewFactoryInterface
+     */
+    private $imageViewFactory;
+
+    /**
+     * @param ImageViewFactoryInterface $imageViewFactory
+     */
+    public function __construct(ImageViewFactoryInterface $imageViewFactory)
+    {
+        $this->imageViewFactory = $imageViewFactory;
```
pjedrzejewski: 👍

PR #109, opened 2017-05-13
[Cart] Extract common summary action

lchrusciel:
Checkout summary and cart summary are the same at the moment.


pjedrzejewski:
👍

PR #125, opened 2017-05-19
[Product] Refactor taxon view

### LINKED ISSUE #120: Change the structure of "taxons" field on the product view
Initially, we only used only taxon ``codes`` in the product response, but we miss some data to generate the breadcrumb for example.  We should include full taxon info, but not the entire tree, just nodes.

lchrusciel:
Resolves #120


pjedrzejewski:
👍

pjedrzejewski:
@letanure @klederson Could you guys confirm this structure is correct, please?

PR #129, opened 2017-05-23
[Cart] Pickup action validation

### LINKED ISSUE #117: Call to a member function getBaseCurrency() on null (Missing validation)
Happens on Cart pickup and add endpoints. Most likely we lack validation if the channel actually exists.

lchrusciel:
Resolves #117


pjedrzejewski:
We should think about how to handle the translation of validation error messages, but looks good so far!

lchrusciel:
@pjedrzejewski I have an idea for that. `ValidationErrorViewFactory` will have injected translator and a requested locale will be passed as a second argument. But I postponed this feature for later.

PR #136, opened 2017-05-26
[Cart] Extract and validate change item quantity action

lchrusciel:
Based on #142

### DIFF
--- a/spec/Command/ChangeItemQuantitySpec.php
+++ b/spec/Command/ChangeItemQuantitySpec.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace spec\Sylius\ShopApiPlugin\Command;
+
+use Sylius\ShopApiPlugin\Command\ChangeItemQuantity;
+use PhpSpec\ObjectBehavior;
+
+final class ChangeItemQuantitySpec extends ObjectBehavior
+{
+    function let()
+    {
+        $this->beConstructedWith('ORDERTOKEN', 1, 5);
+    }
+
+    function it_is_initializable()
+    {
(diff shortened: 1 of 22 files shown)


pjedrzejewski:
Requires rebasing.

spec/Command/ChangeItemQuantitySpec.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        $this->itemIdentifier()->shouldReturn(1);
+    }
+
+    function it_has_quantity()
+    {
+        $this->quantity()->shouldReturn(5);
+    }
+
+    function it_throws_an_exception_if_order_token_is_not_a_string()
```
pamil: Can't we use PHP 7.0 typehints?

spec/Handler/ChangeItemQuantityHandlerSpec.php
```diff
@@ -0,0 +1,87 @@
...
+        $this->beConstructedWith($orderItemRepository, $orderRepository, $orderItemModifier, $orderProcessor);
+    }
+
+    function it_is_initializable()
+    {
+        $this->shouldHaveType(ChangeItemQuantityHandler::class);
+    }
+
+    function it_handles_putting_new_item_to_cart(
+        OrderInterface $order,
```
pamil: Why don't we keep the `let` order of arguments?

spec/Command/ChangeItemQuantitySpec.php
lchrusciel: This code has been written before support for php 70 was introduced. Can be done in the next iteration

pamil:
👍

PR #168, opened 2017-06-22
[Product] [Taxon] Fix product taxon structure

lchrusciel:


src/Factory/ProductViewFactory.php
```diff
@@ -86,13 +86,14 @@ private function getTaxonWithAncestors(TaxonInterface $taxon, $locale): TaxonVie
     {
         $currentTaxonView = $this->taxonViewFactory->create($taxon, $locale);
 
-        if (null === $taxon->getParent()) {
-            return $currentTaxonView;
-        }
+        while (null !== $taxon->getParent()) {
+            $taxon = $taxon->getParent();
```
Arminek: `$taxon` -> `$parentTaxon`

lchrusciel: I would have to assign $parentTaxon to $taxon again at the bottom of while... Not sure if it is worthy.

PR #172, opened 2017-07-03
[Taxons] New taxon details

lchrusciel:


doc/swagger.yml
```diff
@@ -579,24 +579,24 @@ paths:
             items:
               type: "object"
               $ref: "#/definitions/Taxon"
-  /taxons/{slug}:
+  /taxons/{code}:
```
pamil: What if code contains something like `co/de?lol=xd#boom`?

src/Controller/Taxon/ShowTaxonTreeAction.php
```diff
@@ -33,46 +29,30 @@
...
+     * @var string
      */
+    private $fallback;
+
     public function __construct(
         TaxonRepositoryInterface $taxonRepository,
         ViewHandlerInterface $viewHandler,
-        TaxonViewFactoryInterface $taxonViewFactory
+        TaxonViewFactoryInterface $taxonViewFactory,
+        string $fallback
```
pamil: `$fallaback` -> `$fallbackLocale`?

doc/swagger.yml
lchrusciel: It will handle such a code unless it will contain a `/` character

PR #173, opened 2017-07-04
[Product] New way of taxon representation

lchrusciel:


doc/swagger.yml
```diff
@@ -999,6 +997,17 @@ definitions:
         type: "array"
         items:
           $ref: '#/definitions/Image'
+  ProductTaxon:
+    type: "object"
+    properties:
+      main:
+        type: "string"
+        example: "MAIN_TAXON_CODE"
+      others:
```
pamil: Do we stay with `taxons.main` and `taxons.others` nomencalture? It might be misleading as `taxons.main` might also be in `taxons.other`. I'll vote for `mainTaxon` and `taxons` instead (just like in vanilla Sylius).

src/View/ProductTaxonView.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Sylius\ShopApiPlugin\View;
+
+class ProductTaxonView
+{
+    /**
+     * @var string
```
pamil: `string|null` / `?string`

src/View/ProductTaxonView.php
```diff
@@ -0,0 +1,16 @@
...
+     */
+    public $main;
+
+    /**
+     * @var array
```
pamil: `array|string[]`?

src/View/ProductTaxonView.php
```diff
@@ -0,0 +1,16 @@
...
+     */
+    public $others;
```
pamil: `public $others = [];`?

src/View/ProductView.php
```diff
@@ -30,9 +30,9 @@ class ProductView
     public $averageRating;
 
     /**
-     * @var array
+     * @var ProductTaxonView
      */
-    public $taxons = [];
+    public $taxons;
```
pamil: Shouldn't we create an instance of ProductTaxonView in the constructor of this one?

doc/swagger.yml
lchrusciel: I'm not sure if we really should go 1:1 with this similar to vanilla Sylius. What you are saying makes sense, but on the other hand current structure is more readable for me. I would leave it as it is but let's see what does the others think about it. /cc @klederson @pjedrzejewski

pamil: The main thing which disturbs me is that `others` part, let's wait for somebody's else insight :)

PR #175, opened 2017-07-05
[Cart] Present cart discounts

lchrusciel:


pjedrzejewski:
We now have ``discounts`` and ``applied_promotions``, perhaps we should unify this?

lchrusciel:
`discounts` are cart promotion(doesn't matter if applied per item or on the whole cart) and `applied_promotions` are the catalog promotion applied on an item. Maybe we should go with `cartDiscounts` and `catalogDiscounts`? /cc @klederson

src/Factory/CartViewFactory.php
```diff
@@ -82,6 +69,17 @@ public function create(OrderInterface $cart, $localeCode)
             $cartView->payments[] = $this->paymentViewFactory->create($payment, $localeCode);
         }
 
+        $cartDiscounts = [];
```
pamil: `/** @var AdjustmentView[] $cartDiscounts */` before this line?

src/View/CartSummaryView.php
```diff
@@ -4,60 +4,41 @@
...
-     * @var string
-     */
+    /** @var string */
     public $checkoutState;
 
-    /**
-     * @var array
-     */
-    public $items = array();
+    /** @var array|ItemView[[ */
```
pamil: `[[` -> `[]`

PR #178, opened 2017-07-05
[Cart] Minor fixes

lchrusciel:


src/Query/Carts.php
```diff
@@ -34,7 +34,7 @@ public function __construct(
         $this->cartViewFactory = $cartViewFactory;
     }
 
-    public function findByToken(?string $orderToken): CartSummaryView
+    public function findByToken(string $orderToken): CartSummaryView
```
lchrusciel: `findByToken` -> `findOneByToken`

src/Query/Carts.php
```diff
@@ -10,7 +10,7 @@
 use Sylius\ShopApiPlugin\View\CartSummaryView;
 use Webmozart\Assert\Assert;
 
-final class CartQuery implements CartQueryInterface
+final class Carts implements CartsInterface
```
pjedrzejewski: What is the rationale behind such naming? (Not saying it is wrong, I just want to understand.)

lchrusciel: I'm just trying to find a best one. From the DDD perspective, we should have most meaningful words for classes. But `Carts` doesn't convince me either. I will probably move back to CartQuery, but PR has some other fixes, so it still makes sense ;)

pamil: From my POV they're nothing else but repositories, query does not find anything on its own.

src/Query/CartQuery.php
```diff
@@ -34,7 +34,7 @@ public function __construct(
         $this->cartViewFactory = $cartViewFactory;
     }
 
-    public function findByToken(?string $orderToken): CartSummaryView
+    public function findOneByToken(string $orderToken): CartSummaryView
```
pamil: What about keeping to the following convention?
 - `findSomething` if method returns `Something` or `null`
 - `getSomething` if method always returns `Something`

pamil:
The comment I made might get unnoticed so I'm linking to it here -
 (https://github.com/Sylius/SyliusShopApiPlugin/pull/178#discussion_r125939358).

pamil:
Thanks @lchrusciel! 🎉

PR #179, opened 2017-07-06
[Products] Product endpoints by code

lchrusciel:
Based on #178

PR #180, opened 2017-07-07
[Products] Product endpoints by code

lchrusciel:
Based on #178

Recreated PR #179, but from fork.


src/ViewRepository/ProductCatalogViewRepository.php
```diff
@@ -0,0 +1,147 @@
...
+
+    /**
+     * @param string $localeCode
+     * @param PaginatorDetails $paginatorDetails
+     * @param ChannelInterface $channel
+     * @param TaxonInterface $taxon
+     *
+     * @return PageView
+     */
+    private function findByTaxon(string $localeCode, PaginatorDetails $paginatorDetails, ChannelInterface $channel, TaxonInterface $taxon): PageView
```
pamil: Shouldn't we preserve the same order as in `findByTaxon{Code,Slug}` methods? Taxon, locale, channel, paginator details?

src/ViewRepository/ProductDetailsViewRepository.php
```diff
@@ -0,0 +1,92 @@
...
+        ChannelRepositoryInterface $channelRepository,
+        ProductRepositoryInterface $productRepository,
+        ProductViewFactoryInterface $productViewFactory
+    ) {
+        $this->channelRepository = $channelRepository;
+        $this->productRepository = $productRepository;
+        $this->productViewFactory = $productViewFactory;
+    }
+
+    public function findOneBySlug(string $channelCode, string $productSlug, ?string $localeCode): ProductView
```
pamil: What about using some convention for ordering those parameters? It's always something related to subject, locale code and channel (code).

tests/Responses/Expected/product/product_review_list_page_by_slug_response.json
```diff
@@ -0,0 +1,74 @@
...
+            "author": "oliver@queen.com"
+        },
+        {
+            "title": "Nice mug",
+            "rating": @integer@,
+            "comment": "@string@",
+            "author": "oliver@queen.com"
+        },
+        {
+            "title": "Nice mug",
```
pamil: Lol, that mug's really nice. Or the reviews are counterfeited 🌮

src/Model/PaginatorDetails.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+        return $this->limit;
+    }
+
+    public function page(): int
+    {
+        return $this->page;
+    }
+
+    public function getParameters(): array
```
pamil: Something's wrong here:
 - if we don't use getters here, `addParameter` should be named `addGetParameter`
 - if we use getters here, `page` and `limit` methods should get the prefix

src/Model/PaginatorDetails.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+        return $this->getParameters;
+    }
+
+    public function addParameter(string $key, string $value)
```
pamil: `: void`

pamil: Btw. `$value` can be an array as well.

doc/swagger.yml
```diff
@@ -464,12 +464,60 @@ paths:
...
+              required: false
+              type: "integer"
+              example: 2
+            responses:
+                200:
+                    description: "Paginated product list."
+                    schema:
+                        type: "object"
+                        $ref: "#/definitions/ProductsPage"
+    /products-by-slug/{slug}:
```
lchrusciel: `/products/{code}` is typical rest format. We are not really RESTful, but on the other hand, everybody used to it.

src/Request/AddProductReviewByCodeRequest.php
```diff
@@ -0,0 +1,45 @@
...
+    /** @var string */
+    private $comment;
+
+    /** @var string */
+    private $email;
+
+    public function __construct(Request $request)
+    {
+        $this->slug = $request->attributes->get('code');
+
```
lchrusciel: slug is fetched from attributes, rest from the request. I just split them to improve readability.

src/Controller/Product/ShowProductDetailsByCodeAction.php
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        try {
+            return $this->viewHandler->handle(View::create($this->productCatalog->findOneByCode(
+                $request->query->get('channel'),
+                $request->attributes->get('code'),
+                $request->query->get('locale')
+            ), Response::HTTP_OK));
+        } catch (\InvalidArgumentException $exception) {
+            throw new NotFoundHttpException($exception->getMessage());
```
lchrusciel: I just use Assert library to find proper result. We can open an issue with that to different between regular `InvalidArgumentException` and exceptions which could result in `NotFoundHttpException` in api context.

PR #188, opened 2017-07-14
Make view classes customizable

pamil:
Closes #187. Same as https://github.com/Sylius/SyliusElasticSearchPlugin/pull/7.

### DIFF
--- a/src/Factory/AddressViewFactory.php
+++ b/src/Factory/AddressViewFactory.php
@@ -7,12 +7,21 @@ use Sylius\ShopApiPlugin\View\AddressView;
 
 final class AddressViewFactory implements AddressViewFactoryInterface
 {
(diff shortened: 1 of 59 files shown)


src/Factory/AddressViewFactory.php
```diff
@@ -7,12 +7,21 @@
...
+    public function __construct(string $addressViewClass)
+    {
+        $this->addressViewClass = $addressViewClass;
+    }
+
     /**
      * {@inheritdoc}
      */
-    public function create(AddressInterface $address)
+    public function create(AddressInterface $address): AddressView
```
lchrusciel: We cannot be sure that `AddressView` will be returned. But anyway, it can be fixed in the next iteration.

PR #189, opened 2017-07-14
Remove ProductVariantView::$appliedPromotions

pamil:


pamil:
They're not used here and should be an exclusive part of UrbanaraCatalogPromotionPlugin.

lchrusciel:
👍

PR #196, opened 2017-07-31
[Cart] Adding product to cart multiple times bug fix

Zales0123:
- added functionality of modify existing order item if adding product/variant has been added to cart before
- added some tests for new functionality
- added some missing tests for previous implementations that were not covered

### DIFF
--- a/spec/Modifier/OrderModifierSpec.php
+++ b/spec/Modifier/OrderModifierSpec.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace spec\Sylius\ShopApiPlugin\Modifier;
 
 use Doctrine\Common\Persistence\ObjectManager;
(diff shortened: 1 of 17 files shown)


spec/Modifier/OrderModifierSpec.php
```diff
@@ -0,0 +1,82 @@
+<?php
+
```
lchrusciel: We could go with strict types and void return type

tests/Controller/CartPutItemToCartApiTest.php
```diff
@@ -545,6 +600,37 @@ public function it_adds_a_product_variant_based_on_option_to_the_cart()
...
+        $this->loadFixturesFromFile('shop.yml');
+
+        $token = 'SDAOSLEFNWU35H3QLI5325';
+
+        /** @var CommandBus $bus */
+        $bus = $this->get('tactician.commandbus');
+        $bus->handle(new PickupCart($token, 'WEB_GB'));
+
+        $data =
+            <<<EOT
```
lchrusciel: Something went wrong with the formating

Zales0123:
@lchrusciel fixed

pjedrzejewski:
Nice!

PR #197, opened 2017-08-01
Updating shipping cost after decreasing quantity

Zales0123:
Related to https://github.com/Sylius/Sylius/pull/8352.

The service overriding is just a temporary solution, until we could use the implementation proposed in mentioned PR.


pjedrzejewski:
👍 👍 👍

PR #202, opened 2017-08-15
Always include currency in prices

pamil:
Closes #203, closes #204.

### DIFF
--- a/doc/swagger.yml
+++ b/doc/swagger.yml
@@ -1283,10 +1283,13 @@ definitions:
     Price:
         type: "object"
         properties:
-            total:
+            current:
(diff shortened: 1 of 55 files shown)


pjedrzejewski:
Tests are failing.

lchrusciel:
I'm ok with this change but I'm not sure if it is necessary. We have a currency on a cart and it is not possible to have price in a different currency.

pamil:
@lchrusciel it's also for product variants

pjedrzejewski:
@lchrusciel It makes frontend's life a bit easier because the price structure is consistent.

pjedrzejewski:
Thanks Kamil!

PR #216, opened 2017-09-25
Reimplement removing an item from the cart as a standalone action

pamil:
Adds order processing after an item is removed from the cart.


lchrusciel:
👍

PR #223, opened 2017-10-14
[Taxon] Add missing locale error quering taxon details

lchrusciel:


src/Controller/Taxon/ShowTaxonDetailsAction.php
```diff
@@ -44,6 +46,10 @@ public function __invoke(Request $request): Response
         $code = $request->attributes->get('code');
         $locale = $request->query->get('locale');
 
+        if (null === $locale) {
+            throw new BadRequestHttpException('You haven\'t provide a locale!');
```
venyii: `haven't provided` or `didn't provide`

zmiter:
IMHO, maybe it's better to inject $fallbackLocale variable in the same way it's done in ShowTaxonTreeAction?
I was about to send a PR in that time found out an issue has been already posted here =)

mamazu:
I don't think that it should throw an exception this would be more of a thing for validation or as suggested use a default locale.

tests/Responses/Expected/taxon/missing_locale_exception_response.json
```diff
@@ -0,0 +1,4 @@
+{
+    "code": 400,
+    "message": "You haven't provide a locale!"
```
CoderMaggie: A bit of grammar change: "`You did not provide a locale!"`

```suggestion
    "message": "You did not provide a locale!"
```

lchrusciel:
It is not a good solution. Closing for now.

PR #225, opened 2017-10-14
add composer suggestions

sdleiw:
add composer suggestion nelmio/cors-bundle, 
update README for cors-bundle configuration


lchrusciel:
Thanks @sdleiw!

PR #239, opened 2017-11-09
Get quantity from request as an integer.

ce-bo:
PR does fix this issue: https://github.com/Sylius/SyliusShopApiPlugin/issues/235

### DIFF
--- a/src/Request/ChangeItemQuantityRequest.php
+++ b/src/Request/ChangeItemQuantityRequest.php
@@ -31,7 +31,7 @@ final class ChangeItemQuantityRequest
     {
         $this->token = $request->attributes->get('token');
         $this->id = $request->attributes->get('id');
-        $this->quantity = $request->request->has('quantity') ? $request->request->getInt('quantity') : null;
+        $this->quantity = $request->request->getInt('quantity');
     }
 
     /**
--- a/src/Request/ChangeItemQuantityRequest.php
+++ b/src/Request/ChangeItemQuantityRequest.php
@@ -31,7 +31,7 @@ final class ChangeItemQuantityRequest
     {
         $this->token = $request->attributes->get('token');
         $this->id = $request->attributes->get('id');
-        $this->quantity = $request->request->getInt('quantity', null);
+        $this->quantity = $request->request->has('quantity') ? $request->request->getInt('quantity') : null;
     }
 
     /**
--- a/src/Request/ChangeItemQuantityRequest.php
+++ b/src/Request/ChangeItemQuantityRequest.php
@@ -31,7 +31,7 @@ final class ChangeItemQuantityRequest
     {
         $this->token = $request->attributes->get('token');
         $this->id = $request->attributes->get('id');
-        $this->quantity = $request->request->getInt('quantity');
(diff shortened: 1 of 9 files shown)


lchrusciel:
Hey @ce-bo. Thanks for contribution. I would merge it but the tests are failing :( Can you adjust them as well?

ce-bo:
Hi @lchrusciel, the tests have passed now. Could you merge it please?

src/Request/ChangeItemQuantityRequest.php
```diff
@@ -31,7 +31,7 @@ public function __construct(Request $request)
     {
         $this->token = $request->attributes->get('token');
         $this->id = $request->attributes->get('id');
-        $this->quantity = $request->request->get('quantity');
+        $this->quantity = $request->request->has('quantity') ? $request->request->getInt('quantity') : null;
```
pamil: Quantity cannot be null, it's declared as `int`. `$request->request->getInt('quantity', 1)` instead?

src/Request/PutOptionBasedConfigurableItemToCartRequest.php
```diff
@@ -44,7 +44,7 @@ public static function fromArray(array $item)
 
     public static function fromRequest(Request $request)
     {
-        return new self($request->attributes->get('token'), $request->request->get('productCode'), $request->request->get('options'), $request->request->get('quantity'));
+        return new self($request->attributes->get('token'), $request->request->get('productCode'), $request->request->get('options'), $request->request->has('quantity') ? $request->request->getInt('quantity') : null);
```
pamil: Same here and so on.

src/Request/ChangeItemQuantityRequest.php
lchrusciel: `$request->request->getInt('quantity', 0)`? This should trigger validation error. Question is, what should be a default behaviour. 1 as a default would be more flexible for frontends...

pamil: If we allow not to pass quantity then it makes sense to use 1 as the default value. If we require quantity then we should throw an exception here if it does not exist.

ce-bo: The decision is up to you.

lchrusciel: Let's add `1` as default. But we need to mention it in docs.

ce-bo: Can you adjust it, please?

lchrusciel: sure, but it will take some time

pjedrzejewski:
Hi @ce-bo, do you plan to continue on this PR? :)

ce-bo:
Hi @pjedrzejewski, unfortunately not. I already discussed with @lchrusciel that he would rather do it.

lchrusciel:
Thanks @ce-bo for all your effort! And welcome to our community! 🎉

ce-bo:
@lchrusciel thanks for finalizing!

lchrusciel:
I will make a release this week

PR #240, opened 2017-11-16
update nelmio_cors config,

sdleiw:
add allowed header authorization and allowed method options

need to permit cors preflight options, otherwise the JWT token could not be sent along the request like `/shop-api/me`, get always an 'Unauthorized header' error


lchrusciel:
Thanks @sdleiw!

PR #246, opened 2018-01-10
Fix an issue in the configuration example which causes error in loading checkout page.

Hailong:
The CheckoutResolver should only be applied to the subpath of checkout, it would causes error in loading the root checkout path if enabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ Rest of the command are executed inside `project` folder.
 
     sylius_shop:
         checkout_resolver:
-            pattern: "%sylius.security.shop_regex%/checkout/"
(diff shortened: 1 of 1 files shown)


lchrusciel:
Thanks Hailong! And welcome to our community! :) 

Are you willing to provide some tests to avoid regression?

Hailong:
Thank you @lchrusciel , Sylius is a great framework!! I'm actually implementing an eCommerce application based on Sylius and the Shop API, so I guess it will cover all the basic scenarios.

I have also enabled the LexikJWTAuthenticationBundle for Shop API, and added a custom authentication provider to enable a third-party authentication. Now I'm exploring the cart and order workflows...

PR #247, opened 2018-01-12
New Feature: populate the token parameter automatically when authentication is enabled for the APIs.

Hailong:
The idea is to find the cart automatically for logged in users. The key of this change is to assign a token value to the user's cart if it's not there, and use this value to populate the "token" parameter for any following requests.

Therefore, we don't have to pass the token value as a parameter explicitly. I have also added the instruction in the README doc, I'm using this feature for my application. Any feedback is welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,16 @@ In order to check example configuration check
  
 From the test app.
 
+### Cart Token with Authentication
+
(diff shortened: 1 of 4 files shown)


lchrusciel:
The whole point of this plugin is to keep shop stateless. Using CartContext will make it stateful, so it is not a good solution

Hailong:
Thank you @lchrusciel , I got your point. For the current project I'm working on, I will keep it stateful and hopefully it could save some effort in developing the client side. I will see how it goes. I will also keep an eye on how to use the APIs statelessly.

lchrusciel:
Probably listener which will create a token when the user is logged in and maybe some custom route such as `me/cart` would work. 

Anyway, thanks for contribution! :)

Issue #250, opened 2018-01-23
Checkout dedicated events are not triggered


Tetragramat:
Currently no events are triggered while processing checkout, except checkout state machine callbacks.
That means for example emails are not send and whatever else is listening to those checkout related events is not triggered.

See checkout related events doc https://github.com/Sylius/Sylius/blob/master/docs/book/orders/checkout.rst#checkout-related-events

PR #252, opened 2018-01-25
Added dispatching of the checkout dedicated events

Tetragramat:
Checkout dedicated events doc: https://github.com/Sylius/Sylius/blob/master/docs/book/orders/checkout.rst#checkout-related-events

fixes issue #250

### DIFF
--- a/src/Handler/AddressOrderHandler.php
+++ b/src/Handler/AddressOrderHandler.php
@@ -5,12 +5,14 @@ declare(strict_types=1);
 namespace Sylius\ShopApiPlugin\Handler;
 
 use SM\Factory\FactoryInterface;
+use Sylius\Bundle\ResourceBundle\Event\ResourceControllerEvent;
 use Sylius\Component\Core\Factory\AddressFactoryInterface;
 use Sylius\Component\Core\Model\AddressInterface;
 use Sylius\Component\Core\Model\OrderInterface;
 use Sylius\Component\Core\OrderCheckoutTransitions;
 use Sylius\Component\Core\Repository\OrderRepositoryInterface;
 use Sylius\ShopApiPlugin\Command\AddressOrder;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Webmozart\Assert\Assert;
 
 final class AddressOrderHandler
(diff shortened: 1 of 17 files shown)


Tetragramat:
@lchrusciel could someone look at this PR?

Tetragramat:
Thank you for your answer @lchrusciel .
I've looked into those events again.
Events:
`sylius.order.pre_address`
`sylius.order.post_address`
`sylius.order.pre_select_shipping`
`sylius.order.post_select_shipping`
`sylius.order.pre_payment`
`sylius.order.post_payment`
Doesn't do currently anything except redirect in post events.
Event `sylius.order.pre_complete` sets ip and locale into order and checks again consistency of the order. Locale is already set into order by plugin, but ip is not.
Event `sylius.order.post_complete` sends order confirmation email, which is not send by plugin.

So If you do not like triggering the events, then do you think adding submitting the confirmation email would be ok? I think that client ip should be set too.

There is still posibility that when someone makes plugin with listeners listening on checkout events, then that plugin will not work correctly with this plugin because of missing event triggers, but at the same time that plugin may break shop api. So I understand.

lchrusciel:
You are right about last two. Assigning an IP address should be a part of order completion I suppose. With the second one, the answer requires much more work because it seems to be a much better solution to send an event over event bus, so anyone can listen to it(and one of the listeners would send an email). Of course, it should be added to all of the handlers, but I do not have time for it right now :( I could be as simple as dispatching an event inside of handler as the last action, no need of introducing ES. Do you think you can add it to the plugin? 

If no, as a shortcut we can just explicitly call an email service from the controller which dispatch complete command.

Tetragramat:
@lchrusciel I got time now, so I can do dispatching of that event. There is a litle problem. I'm not sure I understand you corrently. What you're asking for is dispatching of an event as the last action in every handler and after that make listener on one of those events which will send an email?
I do not want to redo it multiple times, because I might have no time again like in the last two weeks.

src/Handler/AddressOrderHandler.php
```diff
@@ -81,6 +91,10 @@ public function handle(AddressOrder $addressOrder)
         $order->setShippingAddress($shippingAddress);
         $order->setBillingAddress($billingAddress);
 
+        $this->eventDispatcher->dispatch('sylius.order.pre_address', new ResourceControllerEvent($order));
```
lchrusciel: Missing specs for dispatching commands

Tetragramat: Sorry, this is second time I saw phpspec and I'm unable to figure to how to spec this line.

mamazu:
Is there an update to this?

Tetragramat:
@mamazu I do not plan to continue working on this PR. Anyone can take it and finish it.

mamazu:
Okay, I created #312 but thanks for the foundations.

Zales0123:
Can be closed due to #312

PR #253, opened 2018-02-07
Use Sylius v1.1.0-RC

Zales0123:


composer.json
```diff
@@ -6,7 +6,7 @@
     "require": {
         "php": "^7.1",
 
-        "sylius/sylius": "^1.0",
+        "sylius/sylius": "^1.1",
```
lchrusciel: Is there any reason to restrict version below 1.1?

Zales0123: Well, no 😄

lchrusciel:
Thanks Mateusz Zalewski! 🐋

pamil:
Mateusz Zalewski is great, although we could run tests for both Sylius 1.0 and 1.1 (same as we do with Symfony in Sylius).

Zales0123:
That's true Kamil Kokot! It should be definitely done for this repository. Nice suggestion!

pamil:
All right, I've implemented it in #254. Mateusz Zalewski and Łukasz Chruściel, you might want to check that out!

PR #262, opened 2018-03-09
Added ProductAttributeValue serialization for select attribute types

### LINKED ISSUE #260: Product Attribute Type Select doesn't properly translate (Sylius 1.1)
Likely related to https://github.com/Sylius/Sylius/pull/8766#issue-144617894

When using the shop API, this bundle does not return product select-type attribute values with their proper translations. It instead returns the unique ID of the translation. For example, the API returns as the attribute's value: `["bfbfe392-8cc8-601f-2da3-12106c1fde98"]`, instead of `["Purple"]`. #

gorkalaucirica:
Fixes #260

### DIFF
--- a/src/Factory/Product/ProductAttributeValueViewFactory.php
+++ b/src/Factory/Product/ProductAttributeValueViewFactory.php
@@ -47,7 +47,7 @@ final class ProductAttributeValueViewFactory implements ProductAttributeValueVie
 
         $choices = $configuration['choices'] ?? null;
         $value = $productAttributeValue->getValue()[0] ?? null;
-        if($choices === null || $value === null) {
+        if ($choices === null || $value === null) {
             return null;
         }
 
--- a/src/Factory/Product/ProductAttributeValueViewFactory.php
+++ b/src/Factory/Product/ProductAttributeValueViewFactory.php
@@ -25,7 +25,12 @@ final class ProductAttributeValueViewFactory implements ProductAttributeValueVie
         $productAttributeValueView = new $this->productAttributeValueViewClass();
 
         $productAttributeValueView->code = $productAttributeValue->getCode();
-        $productAttributeValueView->value = $productAttributeValue->getValue();
+
+        if ($productAttributeValue->getType() === 'select') {
+            $productAttributeValueView->value = $this->resolveSelectAttribute($productAttributeValue);
+        } else {
+            $productAttributeValueView->value = $productAttributeValue->getValue();
+        }
 
         $productAttribute = $productAttributeValue->getAttribute();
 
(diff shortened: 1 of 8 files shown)


src/Factory/ProductAttributeValueViewFactory.php
```diff
@@ -37,4 +42,19 @@ public function create(ProductAttributeValueInterface $productAttributeValue, st
...
         return $productAttributeValueView;
     }
+
+    private function resolveSelectAttribute(ProductAttributeValueInterface $productAttributeValue)
+    {
+        $configuration = $productAttributeValue->getAttribute()->getConfiguration();
+
+        if (!isset($configuration['choices'])
+            || count($configuration['choices']) === 0
+            || count($productAttributeValue->getValue()) === 0
```
lchrusciel: This condition fails the build. 

Does all of them are required and covered by tests?

lchrusciel:
@Zales0123 @GSadee can you check if we need to cover anything else or is it good2go?

gorkalaucirica:
@lchrusciel The test was incorrect. `$productAttributeValue->getValue()` returns an array instead of an string.

In case the PR needs more changes just let me know

lchrusciel:
If you could provide one integration test with PHPUnit that would be awesome, but I'm aware that it won't be trivial. I would like to have a confirmation form my colleagues, and we can merge it.

gorkalaucirica:
@lchrusciel I've been checking integration tests, but I need some time to get the whole picture.

If you consider merging it I can try to add coverage to this use case.

Zales0123:
It's a nice feature, so I'd be ok with merging it, but also keeping in mind (and event better on issues list) to cover it better with PHPUnit test as @lchrusciel have already suggested. Anyway, good job 👍 ;)

lchrusciel:
We have two vote's for merging, so I guess we will merge it. But, as was mentioned, we should have an additional unit test. @gorkalaucirica will you give it a try?

gorkalaucirica:
@lchrusciel @Zales0123 Integration tests added, it was easier than expected 💪 .

Those tests are enough?

lchrusciel:
@gorkalaucirica I will iterate one more time over this PR in the nearest future. But first of all, I need to publish a stable release of this plugin

gorkalaucirica:
@lchrusciel Are you releasing another beta or 1.0.0?

This is a BC break in the returned values from the API, so if it's the later will have to wait until 2.0.

mamazu:
Thanks for contributing to the ShopApiPlugin and welcome to the community.

The base of this pull-request was changed, you need fetch and reset your local branch
execute git fetch origin && git reset "feature/product-attribute-view-select" to update your local branch. Feel free to ask for assistance when you get stuck. +1

src/Factory/Product/ProductAttributeValueViewFactory.php
```diff
@@ -25,7 +25,12 @@ public function create(ProductAttributeValueInterface $productAttributeValue, st
         $productAttributeValueView = new $this->productAttributeValueViewClass();
 
         $productAttributeValueView->code = $productAttributeValue->getCode();
-        $productAttributeValueView->value = $productAttributeValue->getValue();
+
+        if ($productAttributeValue->getType() === 'select') {
```
lchrusciel: This `if` could be hidden inside of `resolveAttributeValue`

lchrusciel:
Thanks, Gorka! :1st_place_medal:

PR #276, opened 2018-06-08
Drop support for Sylius 1.0

pamil:
Sylius 1.0 is about to stop being supported in a few days (only receiving security fixes).


lchrusciel:
Thanks Kamil! 🎉

PR #281, opened 2018-06-08
[Travis] Test against different symfony versions

lchrusciel:


Zales0123:
I believe it's already resolved https://github.com/Sylius/ShopApiPlugin/blob/master/.travis.yml#L12 :)

PR #284, opened 2018-08-02
Cleanup and normalize the codebase

bitbager:
Hi @lchrusciel!

I've seen that there are a few coding standards applied to the plugin. I cleaned up it a little. Such great plugin deserves a nice CS 🙂 

Bests,
Nick

### DIFF
--- a/src/Command/AddCoupon.php
+++ b/src/Command/AddCoupon.php
@@ -8,20 +8,12 @@ use Webmozart\Assert\Assert;
 
 final class AddCoupon
 {
-    /**
-     * @var string
-     */
+    /** @var string */
     private $orderToken;
 
-    /**
-     * @var string
-     */
+    /** @var string */
     private $couponCode;
 
-    /**
-     * @param string $orderToken
-     * @param string $couponCode
-     */
     public function __construct($orderToken, $couponCode)
     {
         Assert::allString([$orderToken, $couponCode]);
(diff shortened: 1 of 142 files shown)


lchrusciel:
Thank you very much Mikołaj, but I would prefer to add some EasyCodingStandard also so we can keep the code clean :)

bitbager:
I will clean it up today :)

src/Command/PutOptionBasedConfigurableItemToCart.php
```diff
@@ -48,33 +40,21 @@ public function __construct($orderToken, $product, array $options, $quantity)
...
-     */
     public function orderToken()
     {
         return $this->orderToken;
     }
 
-    /**
-     * @return string
-     */
     public function product()
```
bitbager: @lchrusciel, shouldn't these methods also include return type?

lchrusciel: That's why we need ECS and PHPStan together

src/Command/RemoveCoupon.php
```diff
@@ -8,24 +8,16 @@
...
-    /**
-     * @var string
-     */
+    /** @var string */
     private $orderToken;
 
-    /**
-     * @param string $orderToken
-     */
     public function __construct($orderToken)
```
lchrusciel: Also here we should add typehint

src/Command/AddCoupon.php
```diff
@@ -8,38 +8,24 @@
...
+    /** @var string */
     private $couponCode;
 
-    /**
-     * @param string $orderToken
-     * @param string $couponCode
-     */
     public function __construct($orderToken, $couponCode)
     {
         Assert::allString([$orderToken, $couponCode]);
```
lchrusciel: Every assertion from the constructor should be moved to scalar type hints(if it is possible thus)

lchrusciel:
Closing in favour of #307

PR #229, opened 2017-10-26
[RFC] Address book

jseparovic1:
* [X] GET /address-book
* [X] POST /address-book
* [x] PUT /address-book/{id}
* [X] DELETE /address-book/{id}
* [x] PATCH /address-book/{id}/default

* [x] Add tests
* [x] Add translation labels

PR #285, opened 2018-08-10
Adding an Address book endpoin in the API

mamazu:
This pull request adds a new route to the API `/me/address` that shows the addresses of the currently logged in user.

And some minor fixes.


mamazu:
duplicates #229

PR #286, opened 2018-08-16
Removed unused cart controller

mamazu:
I found an old controller that is not used anymore. So we can safely remove it.

### DIFF
--- a/src/Controller/CartController.php
+++ b/src/Controller/CartController.php
@@ -29,39 +29,7 @@ final class CartController extends Controller
      * @param Request $request
      *
      * @return Response
-     */
-    public function removeItemAction(Request $request)
-    {
-        /** @var OrderRepositoryInterface $cartRepository */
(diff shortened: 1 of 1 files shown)


pamil:
Looks like it is used indeed - https://travis-ci.org/Sylius/SyliusShopApiPlugin/jobs/418205962#L818.

mamazu:
Ok, it looks odd. Maybe we should refactor it.

lchrusciel:
Remove item should be removed for 100%. Delivery cost estimation should be saved, but I'm not sure if we need it for 1.0. 

In the best scenario, we can move shipping price estimation to a better place and adopt new controller implementation.

mamazu:
That's what I thought too.

pamil:
Thank you, Maximilian! 🎉

mamazu:
Your welcome.

Issue #263, opened 2018-03-20
Naming convention


bitbager:
Currently, the plugin does not follow official Sylius plugin naming conventions. I believe this must be changed before the stable release.

mamazu:
What do we still need to do here?

lchrusciel:
#404 should fixed this

PR #290, opened 2018-08-20
Changing plugin to the right naming convention

mamazu:
Making the plugin compatible with the correct naming convention. This requires the plugin to be renamed so you might also need to update the packagist information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Rest of the command are executed inside `project` folder.
             $bundles = [
                 // ...
     
-                new \Sylius\ShopApiPlugin\ShopApiPlugin(),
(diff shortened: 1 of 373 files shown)


lchrusciel:
Thank you very much! But according to rest of our plugins (e.g. https://github.com/Sylius/RefundPlugin) we should rather drop the `Sylius` from the repo name than changing all namespaces. 

/cc @pamil

pamil:
We decided not to apply that naming convention in Sylius' plugins, as it would result in repeating `Sylius` both in repository name and organisation name. I guess we should update the convention to explicitly say it's for 3rd party plugins.

mamazu:
So then the merge request is closed? And ticket #263 is also invalid?

lchrusciel:
Yup, lets close both

Issue #296, opened 2018-08-21
Returning more precise error codes


mamazu:
Currently the endpoints that require a user, throw an exception if there is no user token in the token storage (aka. no user logged in). This in turn results in a 500 Internal Server Error Response which should only be used if there is really something wrong with the server.
I would suggest in this case to return a 401 like for example with the get method on the `/me` endpoint and now the address book endpoints as well.
What do you think?

PR #301, opened 2018-08-27
More precise error messages

mamazu:
As I have already mentioned (#296) I would suggest to use more precise error codes in the address book api to indicate for example that the user is not logged in. I have also added a PHPSpec test for the `CurrentlyLoggedInUserProvider`


mamazu:
Please rerun the build.

doc/swagger.yml
```diff
@@ -912,14 +912,18 @@ paths:
...
               description: "Successfully created the address"
+              schema:
+                - $ref: "#/definitions/LoggedInCustomerAddressBookAddress"
             400:
               description: "Validation failed"
               schema:
                   $ref: "#/definitions/GeneralError"
+            401:
+              description: "No user is logged in"
             500:
```
lchrusciel: It should be 404, I guess

src/Controller/AddressBook/CreateAddressAction.php
```diff
@@ -39,35 +44,44 @@
     private $validationErrorViewFactory;
 
     /**
-     * @var TokenStorage
+     * @var LoggedInUserProviderInterface
      */
-    private $tokenStorage;
+    private $currentUserProvider;
```
lchrusciel: We should adjust the name of field as well

src/Controller/AddressBook/RemoveAddressAction.php
```diff
@@ -39,36 +41,29 @@
...
     private $currentUserProvider;
 
     /**
-     * @param ViewHandlerInterface $viewHandler
-     * @param ValidatorInterface $validator
-     * @param ValidationErrorViewFactory $validationErrorViewFactory
-     * @param CommandBus $bus
-     * @param TokenStorageInterface $tokenStorage
-     * @param CurrentUserProviderInterface $currentUserProvider
+     * @param ViewHandlerInterface          $viewHandler
```
lchrusciel: :(

tests/Controller/AddressBookShowApiTest.php
```diff
@@ -24,4 +24,16 @@ public function it_shows_address_book()
...
+
+    /**
+     * @test
+     */
+    public function it_returns_an_unauthorized_exception_if_there_is_no_logged_in_user()
+    {
+        $this->client->request('GET', '/shop-api/address-book', [], [], ['ACCEPT' => 'application/json']);
+        $response = $this->client->getResponse();
+
+        $this->assertResponse($response, 'address_book/show_address_book_unauthorized_response', Response::HTTP_UNAUTHORIZED);
```
lchrusciel: Can't we just assert response code?

mamazu: Yes

doc/swagger.yml
mamazu: As you may have noticed I have added authorization symbols to the documentation. And the address book is something that only logged in users can see. If you are not logged in and try to call a route where you need to be logged in I wanted to return a 401 Unauthorized.

src/Controller/AddressBook/RemoveAddressAction.php
lchrusciel: We have resigned from this practice, as it is problematic once you want to add a new argument to docblock. People tend to forget about alignment and even if not, then diffs are less readable. Can you revert these changes?

doc/swagger.yml
lchrusciel: I suppose it is not the scope of this PR anyway?

src/Controller/AddressBook/RemoveAddressAction.php
mamazu: Only manually, but I will.

doc/swagger.yml
lchrusciel: Sorry, I was just confused, that we are handling error with `500` code. But this is totally not part of this PR. I'm just missing some part why, but this is rather general question about current codebase :)

src/Validator/AddressExistsValidator.php
```diff
@@ -18,19 +18,19 @@
...
+     * @var LoggedInUserProviderInterface
      */
     private $currentUserProvider;
 
     /**
      * AddressExistsValidator constructor.
      *
-     * @param AddressRepositoryInterface $addressRepository
-     * @param CurrentUserProviderInterface $currentUserProvider
+     * @param AddressRepositoryInterface    $addressRepository
```
lchrusciel: Should be reversed

src/Validator/AddressExistsValidator.php
```diff
@@ -18,19 +18,19 @@
      */
     private $addressRepository;
     /**
-     * @var CurrentUserProviderInterface
+     * @var LoggedInUserProviderInterface
      */
     private $currentUserProvider;
```
lchrusciel: We should rename this parameter in the whole codebase

src/Request/RemoveAddressRequest.php
```diff
@@ -9,12 +9,20 @@
 final class RemoveAddressRequest
 {
     /**
-     * @var mixed
+     * @var int|string
```
lchrusciel: nice 👍

src/Provider/LoggedInUserProvider.php
```diff
@@ -5,17 +5,17 @@
...
-use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorage;
-use Webmozart\Assert\Assert;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+use Symfony\Component\Security\Core\Exception\TokenNotFoundException;
 
-final class CurrentUserProvider implements CurrentUserProviderInterface
+final class LoggedInUserProvider implements LoggedInUserProviderInterface
 {
     /**
-     * @var TokenStorage
```
lchrusciel: 👍

src/Controller/Customer/UpdateCustomerAction.php
```diff
@@ -77,9 +77,11 @@ public function __invoke(Request $request): Response
         $updateCustomerCommand = $updateCustomerRequest->getCommand();
         $this->bus->handle($updateCustomerCommand);
 
-        return $this->viewHandler->handle(View::create(
+        return $this->viewHandler->handle(
+            View::create(
```
lchrusciel: Should be reverted or arguments of `create` method should be indented

src/Controller/AddressBook/UpdateAddressAction.php
```diff
@@ -77,39 +75,56 @@ public function __construct(
...
+            /** @var AddressInterface $updatedAddress */
+            $updatedAddress = $this->addressRepository->findOneBy(['id' => $id]);
 
-        /** @var AddressInterface $updatedAddress */
-        $updatedAddress = $this->addressRepository->findOneBy(['id' => $id]);
+            $view = View::create(
+                $this->addressBookViewFactory->create($updatedAddress, $user->getCustomer()),
+                Response::HTTP_OK
+            );
+        } else {
```
lchrusciel: Instead of `else` we should use early return approach. Using `return $this->viewHandler->handle(...);` with two different arguments should be easier to understand what is happening here

src/Controller/AddressBook/CreateAddressAction.php
```diff
@@ -83,13 +97,40 @@ public function __invoke(Request $request): Response
...
 
-        $user = $this->currentUserProvider->provide();
+        try {
+            /** @var ShopUserInterface $user */
+            $user = $this->loggedInUserProvider->provide();
+        } catch (TokenNotFoundException $exception) {
+            return $this->viewHandler->handle(View::create(null, Response::HTTP_UNAUTHORIZED));
+        }
+
+        if (($customer = $user->getCustomer()) !== null) {
```
lchrusciel: Not needed doubled brackets

spec/Provider/LoggedInUserProviderSpec.php
```diff
@@ -0,0 +1,46 @@
...
+        $tokenStorage->getToken()->shouldBeCalled()->willReturn($token);
+
+        $this->shouldThrow(TokenNotFoundException::class)->during('provide');
+    }
+
+    function it_returns_the_logged_in_user_if_there_is_one(
+        TokenStorageInterface $tokenStorage,
+        TokenInterface $token,
+        ShopUserInterface $shopUser
+    )
```
lchrusciel: CS

lchrusciel:
Thanks @mamazu! 🐋

PR #303, opened 2018-08-31
Modernized cart controller

### LINKED ISSUE #221: Implement last handlers
Cart controller should be split into two actions and two new handlers should be introduced: `removeItemAction` and `estimateShippingCostAction`

mamazu:
Fixes #221


lchrusciel:
I will not be able to do it today, listed for Monday 👍

src/Command/EstimateShippingCost.php
```diff
@@ -0,0 +1,69 @@
...
+        return $this->countryCode;
+    }
+
+    public function provinceCode(): ?string
+    {
+        return $this->provinceCode;
+    }
+
+    /**
+     * @param int    $value
```
lchrusciel: We do not align parameters vertically anymore

src/Command/EstimateShippingCost.php
```diff
@@ -0,0 +1,69 @@
...
+    {
+        $this->result = [$value, $currencyCode];
+    }
+
+    /**
+     *Returns the array of value, currency code
+     *
+     * @return array
+     */
+    public function getResult()
```
lchrusciel: A command cannot return any value!

src/Controller/Cart/EstimateShippingCostAction.php
```diff
@@ -0,0 +1,94 @@
...
+        if (0 !== count($validationResults)) {
+            return $this->viewHandler->handle(
+                View::create(
+                    $this->validationErrorViewFactory->create($validationResults),
+                    Response::HTTP_BAD_REQUEST
+                )
+            );
+        }
+
+        $command = $estimateShippingCostRequest->getCommand();
```
lchrusciel: Instead of using a command pattern which in unidirectional(data can only go to application), we should use a different approach here. Price estimation is more like a read action depending on the current state of the application. What I would propose is to remove the command part about price estimation, create a service in the place of the handler, and pass the required data as an argument.

mamazu: This sounds like a good idea. I will refactor it accordingly.

src/Command/EstimateShippingCost.php
mamazu: My IDE did that. But I will fix it.

src/Controller/Cart/EstimateShippingCostAction.php
mamazu: So should I create a folder services or what would you suggest? I think refactoring it back into the action would also be a possiblity.

lchrusciel: I guess something like Shipping folder and ShippingPriceEstimator would be fine. Do you have any better idea?

mamazu: Yeah that sounds like a good idea. Maybe later there will be more classes so that the `ShippingPriceEstimator` wouldn't feel so lonely. :D

spec/Shipping/EstimateShippingCostEstimatorSpec.php
```diff
@@ -0,0 +1,98 @@
...
+use Sylius\Component\Core\Model\ShipmentInterface;
+use Sylius\Component\Core\Model\ShippingMethodInterface;
+use Sylius\Component\Core\Repository\OrderRepositoryInterface;
+use Sylius\Component\Registry\ServiceRegistryInterface;
+use Sylius\Component\Resource\Factory\FactoryInterface;
+use Sylius\Component\Shipping\Calculator\CalculatorInterface;
+use Sylius\Component\Shipping\Exception\UnresolvedDefaultShippingMethodException;
+use Sylius\Component\Shipping\Resolver\ShippingMethodsResolverInterface;
+
+final class EstimateShippingCostEstimatorSpec extends ObjectBehavior
```
lchrusciel: **EstimateShippingCostEstimator** seems to be a little bit too specific about the estimation :D Can we go with just **ShippingCostEstimator**?

src/Controller/Cart/EstimateShippingCostAction.php
```diff
@@ -0,0 +1,98 @@
...
+        if (0 !== count($validationResults)) {
+            return $this->viewHandler->handle(
+                View::create(
+                    $this->validationErrorViewFactory->create($validationResults),
+                    Response::HTTP_BAD_REQUEST
+                )
+            );
+        }
+
+        [$price, $currency] = $this->shippingCostEstimator->estimate(
```
lchrusciel: Maybe we can introduce some VO?

src/Controller/Cart/EstimateShippingCostAction.php
```diff
@@ -0,0 +1,98 @@
...
+    ) {
+        $this->viewHandler = $viewHandler;
+        $this->shippingCostEstimator = $shippingCostEstimator;
+        $this->validator = $validator;
+        $this->validationErrorViewFactory = $validationErrorViewFactory;
+        $this->priceViewFactory = $priceViewFactory;
+    }
+
+    /**
+     * @param Request $request
```
lchrusciel: This docblock is redundant

src/Controller/Cart/EstimateShippingCostAction.php
```diff
@@ -0,0 +1,98 @@
...
+     */
+    private $validationErrorViewFactory;
+
+    /**
+     * @var PriceViewFactory
+     */
+    private $priceViewFactory;
+
+    /**
+     * @param ViewHandlerInterface $viewHandler
```
lchrusciel: This docblock is redundant

src/Controller/Cart/EstimateShippingCostAction.php
```diff
@@ -0,0 +1,98 @@
...
+     * @param ValidatorInterface $validator
+     * @param ValidationErrorViewFactoryInterface $validationErrorViewFactory
+     * @param PriceViewFactory $priceViewFactory
+     */
+    public function __construct(
+        ViewHandlerInterface $viewHandler,
+        EstimateShippingCostEstimator $shippingCostEstimator,
+        ValidatorInterface $validator,
+        ValidationErrorViewFactoryInterface $validationErrorViewFactory,
+        PriceViewFactory $priceViewFactory
```
lchrusciel: Shouldn;t we go with an interface here?

src/Controller/Cart/EstimateShippingCostAction.php
```diff
@@ -0,0 +1,98 @@
...
+use Sylius\ShopApiPlugin\Shipping\EstimateShippingCostEstimator;
+use Sylius\ShopApiPlugin\View\EstimatedShippingCostView;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
+
+final class EstimateShippingCostAction
+{
+    /**
+     * @var ViewHandlerInterface
```
lchrusciel: Can you change it to the short syntax? `/** @var ViewHandlerInterface */`

src/Shipping/EstimateShippingCostEstimator.php
```diff
@@ -0,0 +1,100 @@
...
+use Sylius\Component\Core\Model\ShippingMethodInterface;
+use Sylius\Component\Core\Repository\OrderRepositoryInterface;
+use Sylius\Component\Registry\ServiceRegistryInterface;
+use Sylius\Component\Resource\Factory\FactoryInterface;
+use Sylius\Component\Shipping\Calculator\CalculatorInterface;
+use Sylius\Component\Shipping\Exception\UnresolvedDefaultShippingMethodException;
+use Sylius\Component\Shipping\Resolver\ShippingMethodsResolverInterface;
+use Webmozart\Assert\Assert;
+
+final class EstimateShippingCostEstimator
```
lchrusciel: As it is no longer a handler, please, provide an interface as well

src/Controller/Cart/EstimateShippingCostAction.php
lchrusciel: It is still not resolved

src/Shipping/ShippingCost.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\Shipping;
+
+class ShippingCost
```
lchrusciel: as its VO, we could finalizes this class I suppose

src/Validator/CountryExistsValidator.php
```diff
@@ -0,0 +1,43 @@
...
+
+final class CountryExistsValidator extends ConstraintValidator
+{
+    /**
+     * @var RepositoryInterface
+     */
+    private $countryRepository;
+
+    /**
+     * CountryExistsValidator constructor.
```
lchrusciel: `* CountryExistsValidator constructor.` is redundant

src/Validator/CountryExistsValidator.php
```diff
@@ -0,0 +1,43 @@
...
+
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+use Sylius\ShopApiPlugin\Validator\Constraints\AddressExists;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+
+final class CountryExistsValidator extends ConstraintValidator
+{
+    /**
+     * @var RepositoryInterface
```
lchrusciel: Several docblocks are unnecessarily expanded

lchrusciel:
Thanks Maximilian! 🐋 Sorry for such a long feedback loop.

src/Validator/CountryExistsValidator.php
mamazu: Should I fix it?

lchrusciel: If you only can, I will be grateful :D

Issue #238, opened 2017-11-08
Codebase cleanup


lchrusciel:
We should enforce a few good practices inside the plugin:

- [x]  Add strict types declaration to all classes
- [x]  Remove assert in favour of type hints
- [ ]  Remove redundant doc blocks(for constructor for example)
- [ ]  Replace param docblocks with short syntax

mamazu:
Now this should be updated to:
:heavy_check_mark: Added strict types to classes (I checked with PHPStorm)
:heavy_check_mark: Remove assert in favour of typehints
:black_circle: Remove redundant doc blocks
:black_circle:  Replace param docblocks with short syntax (what ever that means)

mamazu:
What do you mean with the short syntax?

lchrusciel:
For example https://github.com/Sylius/ShopApiPlugin/pull/307/files#diff-b7b599a1c07c6c9a437e7a82bb0086f3R17

It is done anyway ;)

mamazu:
Okay, I already started that but I wasn't sure if its the right one. :+1:

PR #304, opened 2018-08-31
Replaced asserts with type hints

mamazu:
First part of the #238

### DIFF
--- a/spec/Command/AddCouponSpec.php
+++ b/spec/Command/AddCouponSpec.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace spec\Sylius\ShopApiPlugin\Command;
 
 use PhpSpec\ObjectBehavior;
+use TypeError;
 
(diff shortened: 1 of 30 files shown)


spec/Command/AddCouponSpec.php
```diff
@@ -27,13 +28,13 @@ public function it_throws_an_exception_if_order_token_is_not_a_string()
     {
         $this->beConstructedWith(new \stdClass(), 'COUPON_CODE');
 
-        $this->shouldThrow(\InvalidArgumentException::class)->duringInstantiation();
+        $this->shouldThrow(TypeError::class)->duringInstantiation();
```
lchrusciel: But I would even think about removal of these tests right now. Strict types and scalar typehints should do the trick :)

lchrusciel:
Thanks @mamazu! 🐋 

I hope you don't mind, that I have fixed the build.

PR #306, opened 2018-08-31
Random minor fixes

lchrusciel:


composer.json
```diff
@@ -24,12 +24,23 @@
...
+        ]
+    },
     "autoload-dev": {
         "psr-4": {
             "Tests\\Sylius\\ShopApiPlugin\\": "tests/"
         },
         "classmap": ["tests/Application/app/AppKernel.php"]
     },
+    "config": {
+        "bin-dir": "bin"
```
pamil: Actually, I'd be a little against this change, Symfony 4 default structure keeps vendor binaries in `vendor/bin/` directory and it's worth to stick to one convention.

PR #312, opened 2018-09-27
[WIP] Added dispatching of the checkout dedicated events

mamazu:
This pull request is based on [Tetragramat's PR](https://github.com/Sylius/ShopApiPlugin/pull/252)

and fixes #252

### DIFF
--- a/src/Handler/ChoosePaymentMethodHandler.php
+++ b/src/Handler/ChoosePaymentMethodHandler.php
@@ -5,12 +5,14 @@ declare(strict_types=1);
 namespace Sylius\ShopApiPlugin\Handler;
 
 use SM\Factory\FactoryInterface;
+use Sylius\Bundle\ResourceBundle\Event\ResourceControllerEvent;
 use Sylius\Component\Core\Model\OrderInterface;
 use Sylius\Component\Core\Model\PaymentMethodInterface;
 use Sylius\Component\Core\OrderCheckoutTransitions;
 use Sylius\Component\Core\Repository\OrderRepositoryInterface;
 use Sylius\Component\Core\Repository\PaymentMethodRepositoryInterface;
 use Sylius\ShopApiPlugin\Command\ChoosePaymentMethod;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Webmozart\Assert\Assert;
 
 final class ChoosePaymentMethodHandler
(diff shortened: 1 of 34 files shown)


src/Handler/ChoosePaymentMethodHandler.php
```diff
@@ -55,6 +62,11 @@ public function handle(ChoosePaymentMethod $choosePaymentMethod)
         $payment = $cart->getPayments()[$choosePaymentMethod->paymentIdentifier()];
 
         $payment->setMethod($paymentMethod);
+
+        $this->eventDispatcher->dispatch('sylius.order.pre_payment', new ResourceControllerEvent($cart));
+
         $stateMachine->apply(OrderCheckoutTransitions::TRANSITION_SELECT_PAYMENT);
+
+        $this->eventDispatcher->dispatch('sylius.order.post_payment', new ResourceControllerEvent($cart));
```
lchrusciel: The problem is much more complicated. The problem is that Handlers itself doesn't know anything about persisting to the database. Flush is done in a middleware. Right now, we will dispatch `pre_` and `post_` events together before flushing, so they are no different than state machine callbacks, where the main advantage of the events was to be sure that something was persisted (so we can send an email to the customer).

tests/Controller/CheckoutShowAvailablePaymentMethodsShopApiTest.php
```diff
@@ -86,6 +88,10 @@ public function it_provides_details_about_available_payment_method()
 
         $token = 'SDAOSLEFNWU35H3QLI5325';
 
+        /** @var RequestStack $requestStack */
+        $requestStack = $this->get('request_stack');
```
lchrusciel: Why do we need it?

src/Handler/ChoosePaymentMethodHandler.php
mamazu: So it would be easier to dispatch those events in the middle ware after flushing? This would eliminate the fact that we need to do it for every route seperately.

lchrusciel: Yes, middleware is the answer, but it has to be generic, and this can be problematic

mamazu: Can't we configure the middleware before injecting it into the command handler

tests/Controller/CheckoutShowAvailablePaymentMethodsShopApiTest.php
mamazu: Those empty requests are there because the `ResourceEvent` needs this. I don't know how to work around this.

PR #314, opened 2018-10-04
Logged in Customer placed orders list endpoint

Zales0123:


mamazu:
It's missing documentation though.

PR #321, opened 2018-10-09
Phpstan (level 2)

mamazu:
Added phpstan and enforced level 2.

The thing with the validators is a problem. Maybe you want to add a specific `ConstraintInterface` that defines this but other than that there is not possibility to guarantee that.


lchrusciel:
Do you mean the problem with validating function?
This config should do the trick:
```neon
parameters:
    ignoreErrors:
        - '/::validate\(\) has parameter \$value with no typehint specified\./'
```

lchrusciel:
Thanks Maximilian! 🐋

PR #325, opened 2018-10-18
Refactored the supported locale code out of the repositories

mamazu:
Seeing that most view repositories share the same code to determine the locale that the factory should use, I refactored this out into a provider class that can (and is) be tested.


mamazu:
@lchrusciel what do you think?

spec/ViewRepository/ProductDetailsViewRepositorySpec.php
```diff
@@ -130,39 +113,15 @@ function it_provides_product_view_by_code_in_given_locale(
...
-        ChannelInterface $channel,
-        LocaleInterface $locale
+        SupportedLocaleProviderInterface $supportedLocaleProvider,
+        ChannelInterface $channel
     ) {
-        $channel->getLocales()->willReturn(new ArrayCollection([$locale->getWrappedObject()]));
-        $locale->getCode()->willReturn('en_GB');
-
         $channelRepository->findOneByCode('WEB_GB')->willReturn($channel);
+        $supportedLocaleProvider->provide('de_DE', $channel);
```
lchrusciel: Missing behaviour description

src/Provider/SupportedLocaleProvider.php
```diff
@@ -0,0 +1,43 @@
...
+            Assert::notNull($localeCode);
+        }
+
+        $this->assertLocaleSupport($localeCode, $channel->getLocales());
+
+        return $localeCode;
+    }
+
+    /**
+     * @param string $localeCode
```
lchrusciel: redundant

src/Provider/SupportedLocaleProvider.php
```diff
@@ -0,0 +1,39 @@
...
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\Provider;
+
+use Doctrine\Common\Collections\Collection;
+use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\Component\Locale\Model\LocaleInterface;
+use Webmozart\Assert\Assert;
+
+class SupportedLocaleProvider implements SupportedLocaleProviderInterface
```
lchrusciel: Can we finalize this class?

spec/Provider/SupportedLocaleProviderSpec.php
```diff
@@ -0,0 +1,92 @@
...
+use InvalidArgumentException;
+use PhpSpec\ObjectBehavior;
+use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\Component\Locale\Model\LocaleInterface;
+use Sylius\ShopApiPlugin\Provider\SupportedLocaleProvider;
+use Sylius\ShopApiPlugin\Provider\SupportedLocaleProviderInterface;
+
+final class SupportedLocaleProviderSpec extends ObjectBehavior
+{
+    function it_is_initializable()
```
lchrusciel: Missing void return types in spec

lchrusciel: I know, that's why I didn't put to much attention to this :)

mamazu: I will take it into account when creating further pull requests. :D

Issue #326, opened 2018-10-22
[RFC] Namespacings


mamazu:
Why don't we have namespaces for the Views and the Handler like we do in the Controller?

mamazu:
So that the commands, handlers and request have the same structure as the actions.

lchrusciel:
should have. I think so, as it will be easier to find appropriate class

mamazu:
Would you be interested in a pull request for that?

lchrusciel:
Sure! :D

lchrusciel:
Thank you Maximilian! :)

PR #335, opened 2018-10-30
Splitting up the factories and the validator configs

mamazu:
Starting to split up the service definitions. The moving of the classes will follow.
refs #326


lchrusciel:
But changing class namespaces should be noted in UPGRADE.md as well

lchrusciel:
Thank Maximillian! 🐋

PR #336, opened 2018-10-30
Splitting up the handlers

mamazu:
Refs #335


src/Resources/config/services/handlers.xml
```diff
@@ -1,187 +1,14 @@
...
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
-        <service id="sylius.shop_api_plugin.handler.address_shipment" class="Sylius\ShopApiPlugin\Handler\AddressOrderHandler">
-            <argument type="service" id="sylius.repository.order"/>
-            <argument type="service" id="sylius.factory.address"/>
-            <argument type="service" id="sm.factory"/>
-            <tag name="tactician.handler" command="Sylius\ShopApiPlugin\Command\AddressOrder"/>
-        </service>
+    <imports>
```
lchrusciel: I have an even better idea. We can use wildcard, so all services from some folder and its subfolder will be loaded.

Can you create a next PR, change this import to:
```xml
    <imports>
        <import resource="services/**/*.xml" />
    </imports>
```
and create a service folder, where all of these definitions will be placed? 

Proposed config should be placed in `src/Resources/config/services.xml`

lchrusciel:
Thank Maximillian! 🐋

src/Resources/config/services/handlers.xml
mamazu: This sounds like a good idea. I will do that in the next pull request as you said.

PR #338, opened 2018-11-07
Autogenerate cart token if not passed for pickup & deprecate the old behaviour

pamil:


mamazu:
You should maybe mention it in the `UPGRADE.md` file that passing a token is deprecated now.

tests/Responses/Expected/cart/empty_response.json
```diff
@@ -1,5 +1,5 @@
 {
-    "tokenValue": "SDAOSLEFNWU35H3QLI5325",
+    "tokenValue": "@string@",
```
lchrusciel: Can we use `@uuid@`?

lchrusciel:
Thanks Kami! 🐋

PR #339, opened 2018-11-07
Splitting up the validator

mamazu:
I have split up the validators as anticipated by the service structure.
The question is though: what about the Constraints? For now I left them where they are.


lchrusciel:
Thanks Maximilian! 🐋

PR #342, opened 2018-11-13
Require channel code in all endpoints

Zales0123:
💥 

To support multi-channeling better, we should include channel code in all plugin requests. This is the first step, the next would be including information about the channel to all commands, to be able to use it within the application.

There is lots of code changed, but the idea is quite straightforward 😄The main work is done in `RequestChannelSubscriber` that checks channel code before each request.

My only concern is how to reflect this change in Swagger docs file? I've put `{channelCode}` in `basePath` parameter, but can it be described or explained anywhere? cc @lchrusciel

### DIFF
--- a/src/Checker/ChannelExistenceChecker.php
+++ b/src/Checker/ChannelExistenceChecker.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\Checker;
+
+use Sylius\Component\Channel\Repository\ChannelRepositoryInterface;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+final class ChannelExistenceChecker implements ChannelExistenceCheckerInterface
+{
+    /** @var ChannelRepositoryInterface */
+    private $channelRepository;
+
+    public function __construct(ChannelRepositoryInterface $channelRepository)
+    {
+        $this->channelRepository = $channelRepository;
+    }
+
+    public function withCode(string $channelCode): void
+    {
+        $channel = $this->channelRepository->findOneByCode($channelCode);
+
+        if (null === $channel) {
+            throw new NotFoundHttpException(sprintf('Channel with code %s has not been found.', $channelCode));
+        }
+    }
+}
--- a/src/EventSubscriber/RequestChannelSubscriber.php
+++ b/src/EventSubscriber/RequestChannelSubscriber.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\EventSubscriber;
+
+use Sylius\ShopApiPlugin\Checker\ChannelExistenceCheckerInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\FilterControllerEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+final class RequestChannelSubscriber implements EventSubscriberInterface
+{
+    /** @var ChannelExistenceCheckerInterface */
+    private $channelExistenceChecker;
+
+    public function __construct(ChannelExistenceCheckerInterface $channelExistenceChecker)
+    {
+        $this->channelExistenceChecker = $channelExistenceChecker;
+    }
+
+    public function checkChannelCode(FilterControllerEvent $event): void
+    {
+        $requestAttributes = $event->getRequest()->attributes;
+
+        if (!$requestAttributes->has('channelCode')) {
+            return;
+        }
+
+        $this->channelExistenceChecker->withCode($requestAttributes->get('channelCode'));
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            KernelEvents::CONTROLLER => 'checkChannelCode',
+        ];
+    }
+}
--- a/tests/Controller/AddressBookCreateAddressApiTest.php
+++ b/tests/Controller/AddressBookCreateAddressApiTest.php
@@ -162,7 +162,7 @@ EOT;
         $this->client->request('POST', '/shop-api/SPACE_KLINGON/address-book', [], [], self::$acceptAndContentTypeHeader, $data);
 
         $response = $this->client->getResponse();
-        $this->assertResponse($response, 'address_book/channel_has_not_been_found_response', Response::HTTP_NOT_FOUND);
+        $this->assertResponse($response, 'channel_has_not_been_found_response', Response::HTTP_NOT_FOUND);
     }
 
 }
--- a/tests/Controller/AddressBookCreateAddressApiTest.php
+++ b/tests/Controller/AddressBookCreateAddressApiTest.php
@@ -23,7 +23,7 @@ final class AddressBookCreateAddressApiTest extends JsonApiTestCase
      */
     public function it_allows_user_to_add_new_address_to_address_book()
     {
-        $this->loadFixturesFromFiles(['customer.yml', 'country.yml']);
+        $this->loadFixturesFromFiles(['channel.yml', 'customer.yml', 'country.yml']);
         $this->logInUser('oliver@queen.com', '123password');
 
         $data =
@@ -64,7 +64,7 @@ EOT;
      */
     public function it_does_not_allow_user_to_add_new_address_to_address_book_without_passing_required_data()
     {
-        $this->loadFixturesFromFiles(['customer.yml', 'country.yml']);
+        $this->loadFixturesFromFiles(['channel.yml', 'customer.yml', 'country.yml']);
         $this->logInUser('oliver@queen.com', '123password');
 
         $data =
@@ -91,7 +91,7 @@ EOT;
      */
     public function it_does_not_allow_user_to_add_new_address_to_address_book_without_passing_correct_country_code()
     {
-        $this->loadFixturesFromFiles(['customer.yml', 'country.yml']);
+        $this->loadFixturesFromFiles(['channel.yml', 'customer.yml', 'country.yml']);
         $this->logInUser('oliver@queen.com', '123password');
 
         $data =
@@ -117,7 +117,7 @@ EOT;
      */
     public function it_does_not_allow_user_to_add_new_address_to_address_book_without_passing_correct_province_code()
     {
-        $this->loadFixturesFromFiles(['customer.yml', 'country.yml']);
+        $this->loadFixturesFromFiles(['channel.yml', 'customer.yml', 'country.yml']);
         $this->logInUser('oliver@queen.com', '123password');
 
         $data =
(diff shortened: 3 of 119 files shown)


src/Checker/ChannelExistenceChecker.php
```diff
@@ -0,0 +1,28 @@
...
+    {
+        $this->channelRepository = $channelRepository;
+    }
+
+    public function withCode(string $channelCode): void
+    {
+        $channel = $this->channelRepository->findOneByCode($channelCode);
+
+        if (null === $channel) {
+            throw new NotFoundHttpException(sprintf('Channel with code %s has not been found.', $channelCode));
```
lchrusciel: Can't we make this implementation decoupled from Http? If not, maybe we should move this class to http namespace

src/EventSubscriber/RequestChannelSubscriber.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        $this->channelExistenceChecker = $channelExistenceChecker;
+    }
+
+    public function checkChannelCode(FilterControllerEvent $event): void
+    {
+        $requestAttributes = $event->getRequest()->attributes;
+
+        if (!$requestAttributes->has('channelCode')) {
+            return;
```
lchrusciel: Is it really true? Maybe we should throw an exception if no channel code found as well

src/EventSubscriber/RequestChannelSubscriber.php
```diff
@@ -0,0 +1,39 @@
...
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\EventSubscriber;
+
+use Sylius\ShopApiPlugin\Checker\ChannelExistenceCheckerInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\FilterControllerEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+final class RequestChannelSubscriber implements EventSubscriberInterface
```
lchrusciel: This class is more like `ensurer`

src/EventSubscriber/RequestChannelSubscriber.php
```diff
@@ -0,0 +1,39 @@
...
+
+    public function checkChannelCode(FilterControllerEvent $event): void
+    {
+        $requestAttributes = $event->getRequest()->attributes;
+
+        if (!$requestAttributes->has('channelCode')) {
+            return;
+        }
+
+        $this->channelExistenceChecker->withCode($requestAttributes->get('channelCode'));
```
lchrusciel: Checker may return false and this class could be moved to http namespace

src/Checker/ChannelExistenceChecker.php
mamazu: We could move this logic into the validators section. Because there is already a `ChannelExists` validator.

src/EventSubscriber/RequestChannelSubscriber.php
Zales0123: Unfortunately, there still could be some routes that do not have a channel code in the URL (for example for logging in). I can of course change that, but it would be a big commitment from us :) Maybe we should take it into the consideration and implement such a feature in separate PR?

src/Checker/ChannelExistenceChecker.php
Zales0123: I vote for changing this for something like `ChannelNotFoundException` and move  `RequestChannelSubscriber` to `Http` namespace, where we can catch it and throw a `NotFoundHttpException` to make the request pretty. I would not use `ChannelExists` as it's a specific Symfony validator and I don't think we should base on this component in such a simple (but important) case.

src/EventSubscriber/RequestChannelSubscriber.php
Zales0123: see https://github.com/Sylius/ShopApiPlugin/pull/342#discussion_r233346633

src/EventSubscriber/RequestChannelSubscriber.php
Zales0123: I wanted to keep the Symfony naming here, as we're using a subscribers mechanism from Symfony itself https://symfony.com/doc/current/event_dispatcher.html#creating-an-event-subscriber

bartoszpietrzak1994: As far as I agree with the will of keeping the Symfony naming, the class name itself seems to be misleading for me.

tests/Controller/AddressBookRemoveAddressApiTest.php
```diff
@@ -57,17 +57,31 @@ public function it_returns_bad_request_exception_if_address_has_not_been_found()
...
+
+    /**
+     * @test
+     */
+    public function it_does_not_allow_user_to_delete_address_in_non_existent_channel()
+    {
+        $this->loadFixturesFromFiles(['channel.yml', 'customer.yml', 'country.yml', 'address.yml']);
+        $this->logInUser('oliver@queen.com', '123password');
+
+        $this->client->request('DELETE', sprintf('/shop-api/SPACE_KLINGON/address-book/1'), [], [], self::$acceptAndContentTypeHeader);
```
bartoszpietrzak1994: sprintf() function call is not needed here

tests/Controller/AddressBookSetDefaultAddressApiTest.php
```diff
@@ -45,4 +45,18 @@ public function it_sets_given_address_as_default()
...
+
+    /**
+     * @test
+     */
+    public function it_does_not_allow_to_set_address_as_default_in_non_existent_channel()
+    {
+        $this->loadFixturesFromFiles(['channel.yml', 'customer.yml', 'country.yml', 'address.yml']);
+        $this->logInUser('oliver@queen.com', '123password');
+
+        $this->client->request('PATCH', sprintf('/shop-api/SPACE_KLINGON/address-book/1/default'), [], [], self::$acceptAndContentTypeHeader);
```
bartoszpietrzak1994: sprintf() function call is not needed here

tests/Controller/CheckoutShowAvailablePaymentMethodsShopApiTest.php
```diff
@@ -39,6 +45,9 @@ public function it_does_not_provide_details_about_available_payment_method_befor
         $this->assertResponse($response, 'checkout/get_available_payment_methods_failed', Response::HTTP_BAD_REQUEST);
     }
 
+    /**
+     * TODO check is it possible (test annotation make it fail)
```
bartoszpietrzak1994: What are further steps regarding this TODO?

tests/Controller/CheckoutShowAvailableShippingMethodsShopApiTest.php
```diff
@@ -61,6 +64,9 @@ public function it_provides_details_about_available_shipping_method()
         $this->assertResponse($response, 'checkout/get_available_shipping_methods', Response::HTTP_OK);
     }
 
+    /**
+     * TODO check is it possible (test annotation make it fail)
```
bartoszpietrzak1994: What are further steps regarding this TODO?

tests/Controller/ProductShowCatalogBySlugApiTest.php
```diff
@@ -54,25 +54,48 @@ public function it_shows_second_page_of_paginated_products_from_some_taxon_by_sl
...
 
-        $this->client->request('GET', '/shop-api/taxon-products-by-slug/brands?channel=WEB_GB&limit=1&page=2', [], [], ['ACCEPT' => 'application/json']);
+        $this->client->request('GET', '/shop-api/WEB_GB/taxon-products-by-slug/brands?limit=1&page=2', [], [], ['ACCEPT' => 'application/json']);
         $response = $this->client->getResponse();
 
         $this->assertResponse($response, 'product/limited_product_list_page_by_slug_response', Response::HTTP_OK);
     }
 
+    /**
+     * TODO check is it possible (test annotation make it fail)
```
bartoszpietrzak1994: What are further steps regarding this TODO?

tests/Controller/CheckoutAddressApiTest.php
```diff
@@ -13,14 +13,52 @@
 {
     private static $acceptAndContentTypeHeader = ['CONTENT_TYPE' => 'application/json', 'ACCEPT' => 'application/json'];
 
+    /**
+     * TODO check is it possible (test annotation make it fail)
```
bartoszpietrzak1994: What are further steps regarding this TODO?

tests/Controller/CheckoutShowAvailablePaymentMethodsShopApiTest.php
Zales0123: I plan to open the issue about that, don't want to do anymore in this already big PR. They were not working anyway (as `@test` annotation was missing)

tests/Controller/AddressBookCreateAddressApiTest.php
```diff
@@ -133,9 +133,35 @@ public function it_does_not_allow_user_to_add_new_address_to_address_book_withou
...
+    /**
+     * @test
+     */
+    public function it_does_not_allow_user_to_add_new_address_to_address_book_in_non_existent_channel()
+    {
+        $this->loadFixturesFromFiles(['channel.yml', 'customer.yml', 'country.yml']);
+        $this->logInUser('oliver@queen.com', '123password');
+
+        $data =
+            <<<EOT
```
bartoszpietrzak1994: Move `<<<EOT` to the beginning of the line

tests/Controller/CartPutItemToCartApiTest.php
```diff
@@ -573,10 +573,36 @@ public function it_increases_quantity_of_existing_item_while_adding_the_same_pro
...
+        $bus = $this->get('tactician.commandbus');
+        $bus->handle(new PickupCart($token, 'WEB_GB'));
+
+        $data =
+<<<EOT
+        {
+            "productCode": "LOGAN_MUG_CODE",
+            "quantity": 3
+        }
+EOT;
```
bartoszpietrzak1994: Adding a blank line here would be a welcome change.

tests/Controller/CartPutItemsToCartApiTest.php
```diff
@@ -127,9 +127,48 @@ public function it_shows_validation_error_for_proper_product()
...
+            {
+                "productCode": "LOGAN_HAT_CODE",
+                "options": {
+                    "HAT_SIZE": "HAT_SIZE_S",
+                    "HAT_COLOR": "HAT_COLOR_RED"
+                }
+            }
+        ]
+EOT;
+        $this->client->request('POST', sprintf('/shop-api/SPACE_KLINGON/carts/%s/multiple-items', $token), [], [], static::$acceptAndContentTypeHeader, $data);
```
bartoszpietrzak1994: Adding a blank line here would be a welcome change.

tests/Controller/CustomerUpdateCustomerApiTest.php
```diff
@@ -97,8 +97,31 @@ public function it_does_not_allow_to_update_customer_without_passing_required_da
...
+<<<EOT
+        {
+            "firstName": "",
+            "lastName": "",
+            "email": "",
+            "gender": "",
+            "phoneNumber": ""
+        }
+EOT;
+        $this->client->request('PUT', '/shop-api/SPACE_KLINGON/me', [], [], self::$contentTypeHeader, $data);
```
bartoszpietrzak1994: Adding a blank line here would be a welcome change.

tests/Controller/ProductAddReviewByCodeApiTest.php
```diff
@@ -18,7 +18,7 @@ public function it_adds_review_to_product()
         $this->loadFixturesFromFiles(['shop.yml']);
 
         $data =
-<<<EOT
+            <<<EOT
```
bartoszpietrzak1994: Move `<<<EOT` to the beginning of the line.

tests/Controller/ProductAddReviewByCodeApiTest.php
```diff
@@ -50,9 +50,32 @@ public function it_adds_review_to_the_product_for_registered_user()
...
+<<<EOT
+        {
+            "channelCode": "WEB_GB",
+            "title": "Awesome product",
+            "rating": 5,
+            "comment": "If I were a mug, I would like to be like this one!",
+            "email": "oliver@example.com"
+        }
+EOT;
+        $this->client->request('POST', '/shop-api/SPACE_KLINGON/products/LOGAN_MUG_CODE/reviews', [], [], self::$acceptAndContentTypeHeader, $data);
```
bartoszpietrzak1994: Adding a blank line here would be a welcome change.

PR #349, opened 2018-11-21
Fixed order addressing filling the database with garbage

JakobTolkemit:
**Before:** Addressing an order would always create two new addresses, even when we just want to change the address of an order, which fills the database with a lot of unused addresses.

**After:** The order addressing handler looks for existing addresses on the order and updates them if possible.

### DIFF
--- a/spec/Handler/Cart/AddressOrderHandlerSpec.php
+++ b/spec/Handler/Cart/AddressOrderHandlerSpec.php
@@ -31,8 +31,72 @@ final class AddressOrderHandlerSpec extends ObjectBehavior
         OrderRepositoryInterface $orderRepository,
         StateMachineInterface $stateMachine
     ): void {
+        $order->getShippingAddress()->willReturn(null);
+        $order->getBillingAddress()->willReturn(null);
+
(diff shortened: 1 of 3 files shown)


mamazu:
@JakobTolkemit Could you fix the merge conflicts and then mention lchrusciel to check it?

lchrusciel:
The PR is checked already. What we need here is one more functional test. Just to check if the order is addressed twice only one address exists in the database.

mamazu:
Is that a thing to test with unit or better with spec?

lchrusciel:
The spec is already there. I would add one unit with it.

mamazu:
I'll see if I can get to it today.

mamazu:
@lchrusciel Jakob implemented it. :D

mamazu:
@JakobTolkemit Could you resolve the conflicts that have occurred with the current master?

mamazu:
Thanks for contributing to the ShopApiPlugin and welcome to the community.

The base of this pull-request was changed, you need fetch and reset your local branch
execute git fetch origin && git reset "fix_order_addressing" to update your local branch. Feel free to ask for assistance when you get stuck. :+1:

mamazu:
Thanks, Jakob! :tada:

PR #351, opened 2018-11-23
Checkout bug

mamazu:
Customer needs to be logged in in order to use the checkout.

### DIFF
--- a/spec/Handler/CompleteOrderHandlerSpec.php
+++ b/spec/Handler/CompleteOrderHandlerSpec.php
@@ -15,7 +15,7 @@ use Sylius\Component\Core\Repository\CustomerRepositoryInterface;
 use Sylius\Component\Core\Repository\OrderRepositoryInterface;
 use Sylius\Component\Resource\Factory\FactoryInterface;
 use Sylius\ShopApiPlugin\Command\CompleteOrder;
-use Sylius\ShopApiPlugin\Exception\NotLoggedInException;
+use Sylius\ShopApiPlugin\Exception\WrongUserException;
 use Sylius\ShopApiPlugin\Provider\LoggedInUserProviderInterface;
 
 final class CompleteOrderHandlerSpec extends ObjectBehavior
@@ -139,7 +139,7 @@ final class CompleteOrderHandlerSpec extends ObjectBehavior
         $order->setCustomer($customer)->shouldNotBeCalled();
         $stateMachine->apply('complete')->shouldNotBeCalled();
 
-        $this->shouldThrow(NotLoggedInException::class)
+        $this->shouldThrow(WrongUserException::class)
              ->during('handle', [new CompleteOrder('ORDERTOKEN', 'example@customer.com', 'Some notes')])
         ;
     }
--- a/spec/Handler/CompleteOrderHandlerSpec.php
+++ b/spec/Handler/CompleteOrderHandlerSpec.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace spec\Sylius\ShopApiPlugin\Handler;
 
 use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
 use SM\Factory\FactoryInterface as StateMachineFactoryInterface;
 use SM\StateMachine\StateMachineInterface;
 use Sylius\Component\Core\Model\CustomerInterface;
@@ -17,6 +18,7 @@ use Sylius\Component\Resource\Factory\FactoryInterface;
 use Sylius\ShopApiPlugin\Command\CompleteOrder;
 use Sylius\ShopApiPlugin\Exception\WrongUserException;
 use Sylius\ShopApiPlugin\Provider\LoggedInUserProviderInterface;
+use Symfony\Component\Security\Core\Exception\TokenNotFoundException;
 
 final class CompleteOrderHandlerSpec extends ObjectBehavior
 {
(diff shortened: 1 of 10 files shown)


spec/Handler/CompleteOrderHandlerSpec.php
```diff
@@ -9,28 +9,66 @@
...
+        ShopUserInterface $shopUser,
         OrderInterface $order,
         OrderRepositoryInterface $orderRepository,
         StateMachineFactoryInterface $stateMachineFactory,
         StateMachineInterface $stateMachine
     ): void {
         $orderRepository->findOneBy(['tokenValue' => 'ORDERTOKEN'])->willReturn($order);
-        $customerProvider->provide('example@customer.com')->willReturn($customer);
+
+        $customerRepository->findOneBy(['email' => 'example@customer.com'])->willReturn($customer);
```
lchrusciel: Do we need to fetch customer from the database? Won't checking just the customer's mail be enough?

spec/Handler/CompleteOrderHandlerSpec.php
```diff
@@ -63,11 +106,42 @@ function it_handles_order_completion_with_notes(
...
+        $shopUser->getCustomer()->willReturn($loggedInCustomer);
+        $loggedInUserProvider->provide()->willReturn($shopUser);
+
+        $stateMachineFactory->get($order, OrderCheckoutTransitions::GRAPH)->willReturn($stateMachine);
+        $stateMachine->can('complete')->willReturn(true);
+
+        $order->setCustomer($customer)->shouldNotBeCalled();
+        $stateMachine->apply('complete')->shouldNotBeCalled();
+
+        $this->shouldThrow(WrongUserException::class)
```
lchrusciel: 👍 but I'm not sure about naming. WDYT about `CheckingOutOnBehalfOfAnotherUserException`?

src/Controller/Checkout/CompleteOrderAction.php
```diff
@@ -21,37 +22,53 @@
...
 
     private function provideUserEmail(Request $request): string
     {
-        $user = $this->tokenStorage->getToken()->getUser();
-
-        if ($user instanceof ShopUserInterface) {
-            return $user->getCustomer()->getEmail();
+        try {
+            return $this->loggedInUserProvider->provide()->getEmail();
+        } catch (TokenNotFoundException $tokenNotFoundException) {
```
lchrusciel: Do we need to base our login on exception? I would prefer to use if statement and add a new method in `loggedInUserProvider`/ allow a null return in `provide`

src/Handler/CompleteOrderHandler.php
```diff
@@ -44,11 +58,34 @@ public function handle(CompleteOrder $completeOrder)
...
+        if ($customer === null) {
+            /** @var CustomerInterface $customer */
+            $customer = $this->customerFactory->createNew();
+            $customer->setEmail($emailAddress);
+
+            return $customer;
+        }
+
+        // If the customer does exist the user has to be logged in with this customer. Otherwise the user is not authorized to complete the checkout
+        $loggedInUser = $this->loggedInUserProvider->provide();
```
lchrusciel: That's not exactly true. Customer may exist without ShopUser on only during first guest checkout. We need to cover:

1. User is logged in - valid
2. User is not logged in and customer exists, but doesn't have an account - valid
3. User is not logged in and customer exists, but has an account - exception, he needs to log in
4. User is logged in but mail is provided - exception

spec/Handler/CompleteOrderHandlerSpec.php
mamazu: We need the customer object later on anyways to add it to the order.

src/Controller/Checkout/CompleteOrderAction.php
mamazu: Well that is how the user provider is implemented. I am sure we could also add another method to the `LoggedInUserProvider` that checks if a user is logged in or not without throwing an exception.

spec/Handler/CompleteOrderHandlerSpec.php
mamazu: A little too descriptive but sure, I can change that. I would have a catchier name in mind something like `IdentityTheftException`

src/Handler/CompleteOrderHandler.php
lchrusciel: That would be the perfect solution

src/Controller/Checkout/CompleteOrderAction.php
lchrusciel: Would be awesome. I don't like base business logic on exceptions

src/Handler/CompleteOrderHandler.php
mamazu: Then I guess we should introduce another route for checkout. Otherwise this might lead to confusion with the documentation.

src/Controller/Checkout/CompleteOrderAction.php
mamazu: I will add that to the interface and then to the `UPDATE.md` file.

spec/Handler/CompleteOrderHandlerSpec.php
mamazu: This makes it possible to use it also in other places.

mamazu:
@lchrusciel Can you have a look at it again?

lchrusciel:
Closing in favour of #365

PR #352, opened 2018-11-25
[WIP] add validation for address orders to prevent exceptions (type hi…

healerz:
…nts) in case of missing properties provided by the client and to stick to the command bus pattern

### DIFF
--- a/src/Request/AddressOrderRequest.php
+++ b/src/Request/AddressOrderRequest.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\Request;
+
+use Sylius\ShopApiPlugin\Command\AddressOrder;
+use Sylius\ShopApiPlugin\Model\Address;
+use Symfony\Component\HttpFoundation\Request;
+
+final class AddressOrderRequest
+{
+    /** @var string|null */
+    private $orderToken;
+
+    /** @var AddressRequest */
+    private $address;
+
+    /** @var AddressRequest */
+    private $billingAddress;
+
+    public function __construct(Request $request)
+    {
+        $this->orderToken = $request->attributes->get('token');
+        $this->address = new AddressRequest($request->request->get('shippingAddress'));
+        $this->billingAddress = new AddressRequest(
+            $request->request->get('billingAddress') ?: $request->request->get('shippingAddress')
+        );
+    }
+
+    public function getCommand(): AddressOrder
+    {
+        return new AddressOrder(
+            $this->orderToken,
+            Address::createFromAddressRequest($this->address),
+            Address::createFromAddressRequest($this->billingAddress)
+        );
+    }
+}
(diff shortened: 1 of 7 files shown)


src/Request/AddressRequest.php
```diff
@@ -0,0 +1,101 @@
...
+    /** @var string|null */
+    private $provinceCode;
+
+    /** @var string|null */
+    private $company;
+
+    /** @var string|null */
+    private $phoneNumber;
+
+    public function __construct($address) {
```
mamazu: Missing a typehint

src/Request/AddressOrderRequest.php
```diff
@@ -0,0 +1,39 @@
...
+    /** @var AddressRequest */
+    private $address;
+
+    /** @var AddressRequest */
+    private $billingAddress;
+
+    public function __construct(Request $request)
+    {
+        $this->orderToken = $request->attributes->get('token');
+        $this->address = new AddressRequest($request->request->get('shippingAddress'));
```
mamazu: We also need to validate that `$request->request->get('shippingAddress')` is not null.

healerz:
just in case the context of this PR is not clear, here is the related issue i created: https://github.com/Sylius/ShopApiPlugin/issues/353

src/Request/AddressRequest.php
healerz: Actually, I didn't type hint the param on purpose. As I see it, the request objects are just loose data holder objects, which need to be validated by the validator anyway and they shouldn't end in an exception in any case.
But you are right, in this case I could type hint it with "?array". On the other hand, if I finish the whole PR, i also need to adjust the code for the Address actions and then i need to modify the Address constructor to something that fits for both cases anyway.

mamazu: Then if `$address` is null you can not use the array access to get properties.

src/Request/AddressOrderRequest.php
mamazu: Or you could add a default to it that it returns an empty array. Like so: `$request->request->get('shippingAddress', [])`

src/Request/AddressRequest.php
healerz: with the `??` operator i can ;)

> php > $foo = null;
php > var_dump($foo['bar'] ?? null);
NULL
php > $foo['bar'] = 'bar';
php > var_dump($foo['bar'] ?? null);
string(3) "bar"

src/Request/AddressOrderRequest.php
healerz: yes i could :) The behaviour is in either way the same

src/Request/AddressRequest.php
mamazu: Ah okay. Didn't expect that to work. :D

healerz: The magic of the `??` operator :D
It's just the shorthand of `isset($foo['bar']) ? $foo['bar] : null`

mamazu:
The idea looks good. This also allows for better error handling in the API client.

mamazu:
If you fix the tests. This feature would be ready for release I'd say.

mamazu:
This should be closed by #453. Isn't it? @healerz

lchrusciel:
Hey Michel, 

please reopen if still relevant :)

PR #355, opened 2018-11-26
Re-done get order by tokenValue instead of ID

kortwotze:
This is a re-do of #347 

It is basically the same commit, but instead of allowing getting an order with ID or tokenValue, this time I switched to only use the tokenValue.

I think it is not good practice to expose the ID of an order, which is not obtainable by using the API in any way. Using the token only should be the way to go, because the token should already be present in all clients using the API and is also obtainable through picking up a cart.

### DIFF
--- a/spec/ViewRepository/PlacedOrderViewRepositorySpec.php
+++ b/spec/ViewRepository/PlacedOrderViewRepositorySpec.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace spec\Sylius\ShopApiPlugin\ViewRepository;
 
 use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
 use Sylius\Component\Core\Model\CustomerInterface;
 use Sylius\Component\Core\Model\OrderInterface;
 use Sylius\Component\Core\OrderCheckoutStates;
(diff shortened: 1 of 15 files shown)


spec/ViewRepository/PlacedOrderViewRepositorySpec.php
```diff
@@ -102,9 +104,6 @@ function it_throws_exception_if_there_is_no_customer_with_given_email(
             ->during('getAllCompletedByCustomerEmail', ['test@example.com'])
         ;
 
-        $this
-            ->shouldThrow(\InvalidArgumentException::class)
-            ->during('getOneCompletedByCustomerEmailAndId', ['test@example.com', 1])
-        ;
+        $orderRepository->findOneBy(Argument::any())->shouldNotBeCalled();
```
bartoszpietrzak1994: I would place this line before following assertions:

```
$this
    ->shouldThrow(\InvalidArgumentException::class)
    ->during('getAllCompletedByCustomerEmail', ['test@example.com'])
;
```

kortwotze: You mean like this?

```
        $customerRepository->findOneBy(['email' => 'test@example.com'])->willReturn(null, null);
        $orderRepository->findOneBy(Argument::any())->shouldNotBeCalled();

        $this
            ->shouldThrow(\InvalidArgumentException::class)
            ->during('getAllCompletedByCustomerEmail', ['test@example.com'])
        ;
```

bartoszpietrzak1994: That's right :)

kortwotze: Will do. I keep learning ;)

PR #357, opened 2018-12-04
ChannelCode taken from url instead of request

bartoszpietrzak1994:


lchrusciel:
Thank you, Bartosz! :1st_place_medal:

Issue #359, opened 2018-12-10
register endpoint does not store email_verification_token


aayaresko:
after a user gets registered via shop-api/register endpoint the 'email_verification_token' field of the user is null. the user cannot confirm his email.

aayaresko:
found what's causing it. another config for tactician overrides the one from the SyliusShopApi

mamazu:
Ah, but that is not a Sylius issue/plugin but something specific to your instance?

aayaresko:
well, with symfony4 we have a case when league_tactician.yaml from config/packages overrides the one for SyliusShopApi.

aayaresko:
I made copy-pasted the config from SyliusShopApi to the config/packages

PR #360, opened 2018-12-10
missing email_verification_token fix (#359)

aayaresko:
issue #359

### DIFF
--- a/src/EventListener/UserRegistrationListener.php
+++ b/src/EventListener/UserRegistrationListener.php
@@ -59,6 +59,7 @@ final class UserRegistrationListener
         }
 
         $this->bus->handle(new GenerateVerificationToken($event->email()));
+        $this->userManager->flush();
         $this->bus->handle(new SendVerificationToken($event->email(), $event->channelCode()));
     }
 }


mamazu:
As far as I know the command bus flushes everything anyways after executing the middle ware.

aayaresko:
okay, I'll check that on my local machine one more time.

aayaresko:
no, sir. the email_verification_token is empty without that line...

aayaresko:
could you please check the registration flow so we both would know that there is no need to update the code of the plugin an everything works as expected

mamazu:
Sure thing.

aayaresko:
> Sure thing.

thanx a lot!

mamazu:
Hm, I can't find where the command service is configured. Could you have a look. The compiler doesn't seem to generate the service definition. Do you have a service definition for "command.bus"?

aayaresko:
no, I don't

aayaresko:
okay, let's see...

aayaresko:
found what's causing it. another config for tactician overrides the one from the SyliusShopApi

aayaresko:
closing the request

PR #362, opened 2018-12-12
Splitting up the view repositories

mamazu:
Moving the view repositories in their own namespaces.

### DIFF
--- a/src/Resources/config/services/controllers.xml
+++ b/src/Resources/config/services/controllers.xml
@@ -2,12 +2,6 @@
 
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <imports>
-        <import resource="actions/address_book.xml"/>
-        <import resource="actions/checkout.xml"/>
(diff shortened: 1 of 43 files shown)


src/Resources/config/services/controllers.xml
```diff
@@ -2,12 +2,6 @@
...
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <imports>
-        <import resource="actions/address_book.xml"/>
-        <import resource="actions/checkout.xml"/>
-        <import resource="actions/cart.xml"/>
-        <import resource="actions/customer.xml"/>
-        <import resource="actions/order.xml"/>
-        <import resource="actions/product.xml"/>
-        <import resource="actions/taxon.xml"/>
+        <import resource="actions/**/*"/>
```
lchrusciel: Can't we just use generic import from services.xml?

mamazu: Sure, I'll try to do that. But some files aren't split up properly.

lchrusciel: That is ok. This PR may be merged, but first I need to make my mind with #351.

lchrusciel:

PR #367, opened 2019-01-09
Splitting up the handlers

mamazu:
Reorganizing the handler into the respective structure.


lchrusciel:
Thank you, @mamazu! :tada:

PR #372, opened 2019-01-16
Upgrade to Sylius ^1.3

pamil:
We need to figure out how to test plugins on Symfony 3.4 and 4.2, but that's the first step allowing for change from Tactician to Messenger in order to unify our message bus used across Sylius packages.


.travis.yml
```diff
@@ -36,11 +16,10 @@ before_install:
 
 install:
     - composer require sylius/sylius:${SYLIUS_VERSION} --no-interaction --no-update
-    - composer require symfony/symfony:${SYMFONY_VERSION} --no-interaction --no-update
     - composer update --no-interaction --prefer-dist
 
 before_script:
-    - (cd tests/Application && bin/console doctrine:schema:create --env=test)
+    - tests/Application/bin/console doctrine:schema:create
```
bartoszpietrzak1994: The build says that this is not working :(

pamil: That's true 😢

Issue #251, opened 2018-01-25
Move swagger documentation out of repository and split specification files


DeRain:
Hello, actually I've done this work but I'm not sure how should I relate this to the plugin. If you could create repository under the Sylius organization based in mine - it would be fine.

Please, check my repo with the specification - it looks much better that one file solution. Also, i've fixed some validation errors for the specification and now it's fully compatible with swaggerhub and we have a option to provide a documentation in UI and run a mock server as well.

**The repo (currently it generates a swagger.json, but it could be easily converted into yaml by the swaggerhub):**
https://github.com/DeRain/SyliusShopApiPlugin-Swagger

**Swaggerhub:**
https://app.swaggerhub.com/apis/DeRain/sylius-shop_api/1.0.0

Zales0123:
Hello Kirill and thank you very much for your effort! You did a wonderful job, indeed.
To be honest, I think we need some time to discuss how we want to handle this issue. Sadly but truly, I think we can't keep documentation in the separate repository (we also abandon this idea for https://github.com/Sylius/Sylius some time ago), on the other hand we can't keep such advanced application in this repo, only for the documentation... However it would be a big shame to not appreciate your hard work on the documentation and to not use your code :) 
Let's keep this issue open and I think that in (I hope so) the nearest future we will take some decisions about it.
Thank you one more time for you work!

DeRain:
@Zales0123 Hello Mateusz and thank you for the response. Sure, ping me, if you will need help in any case with this. As an option - it could be distributed as an npm package with API testing as well using language agnosting framework Dredd (https://github.com/apiaryio/dredd). This tool takes swagger documentation as an input and applied automatic tests for an API according to swagger.

More info here
https://blog.apisyouwonthate.com/keeping-documentation-honest-d9ab5351ddd4

So the basic idea is keep documentation and implementation separated as it helps to uncover outdated documentation and broken features.

Thank you for reading!

lchrusciel:
Are you willing to give a try to set up dredd for this project? Or t least give it a try? I don't have any experience with it, but it would be an awesome feature to have it

DeRain:
Currently, i'm working on the project based on Sylius and when i will have
a chance I will setup Dredd for this project and then I think I will be
able to adapt this for SyliusShopApiPlugin.

Thank you!

mamazu:
Is there any update to this?

lchrusciel:
no :(

[...]

PR #376, opened 2019-01-24
Fixed swagger.yml to match API

dlobato:
Schema is valid now.
There's also some fixes on several object definitions to match current fields on API request/responses.

I like very much the proposed solution on https://github.com/Sylius/ShopApiPlugin/issues/251 but editing the file in one piece seems more simple.

### DIFF
--- a/doc/swagger.yml
+++ b/doc/swagger.yml
@@ -11,1686 +11,1686 @@ info:
 host: "demo.sylius.org"
 basePath: "/shop-api"
 consumes:
-  - "application/json"
+    - "application/json"
 produces:
-  - "application/json"
+    - "application/json"
 tags:
-- name: "cart"
-  description: "All actions related to cart management."
-  externalDocs:
-      description: "Find out more"
-      url: "http://docs.sylius.org/en/latest/book/orders/index.html"
-- name: "products"
-  description: "Show product catalog and add product reviews."
-  externalDocs:
-      description: "Find out more"
-      url: "http://docs.sylius.org/en/latest/book/products/index.html"
-- name: "taxons"
-  description: "Show taxon tree"
-  externalDocs:
-      description: "Find out more"
-      url: "http://docs.sylius.org/en/latest/book/products/taxons.html"
-- name: "checkout"
-  description: "All actions related to checkout fulfillment. It is important, to execute them in given order (address, choose shipment, choose payment and complete)."
-  externalDocs:
-      description: "Find out more"
-      url: "http://docs.sylius.org/en/latest/book/orders/checkout.html"
-- name: "order"
-  description: "Showing the order information"
-  externalDocs:
-    description: "Find out more"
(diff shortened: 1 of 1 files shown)


doc/swagger.yml
```diff
@@ -9,7 +9,11 @@ info:
         name: "MIT"
         url: "https://opensource.org/licenses/MIT"
 host: "demo.sylius.org"
-basePath: "/shop-api/{channelCode}"
+basePath: "/shop-api"
```
mamazu: Thanks for clarifying that.

dlobato: yep, according to the specification: 

> The basePath does not support path templating.

dlobato:
@mamazu I did test most of the endpoints and so far I did not find any mismatch on the response codes. Anyway we are implementing a client right now, so we will very likely spot problems.

doc/swagger.yml
mamazu: And there are some routes that don't follow that pattern I believe like the customer endpoints.

mamazu:
Normally the response codes in the Actions would give it away but I am sure you'll find those errors as well.

mamazu:
The current version of the documentation has different levels of indentation in different parts. But 2 spaces are the convention for yml.

dlobato:
@lchrusciel No reason for changing. But because there was 4 and 2 on different places, I just normalized to 2.  There's still 4 spaces on some places... I'll fix it. I don't have any preference. What do you prefer?

dlobato:
2 spaces => https://pastebin.com/dKZKPC5s
4 spaces => https://pastebin.com/ufQQX01F

lchrusciel:
Oh, the main problem was, that it was hard to comprehend changes in the file, what has been changed. In this situation, it will not help much. Nevertheless, I would prefer 4, if you don't mind changing it.

doc/swagger.yml
```diff
@@ -208,33 +182,26 @@ paths:
...
-          in: "path"
-          description: "Cart identifier."
-          required: true
-          type: "string"
-        - name: "identifier"
-          in: "path"
-          description: "Identifier of a specific item. Can be found in the cart summary."
-          required: true
-          type: "string"
+            -   $ref: "#/parameters/ChannelCode"
```
lchrusciel: Should there be so many spaces after `-` and before `$`

dlobato: phpstorm indent this way... I don't think there's any standard way to format yml.
For instance, https://jsonformatter.org/yaml-formatter will do:
```
...
/{channelCode}/product/{code}/reviews:
    parameters:
        -
            $ref: '#/parameters/ChannelCode'
        -
            name: code
            in: path
            description: 'Code of expected product.'
            required: true
            type: string
...
```

lchrusciel:
Thank you, David! :1st_place_medal:

PR #380, opened 2019-01-25
[POC] New command providers

lchrusciel:


mamazu:
I have rebased the pull request to the current master and made a minor change. The `DropCartCommandProvider` is now extendable. This makes it much more simple to replace the `Request` object that is being generated as well as the `command` object that is generated. This makes it much extending this structure much easier and not having to override the controller itself when changing the command and the request.

@lchrusciel What do you think. Sorry for hijacking your pull request. :D

src/CommandProvider/CommandProviderInterface.php
```diff
@@ -0,0 +1,15 @@
...
+namespace Sylius\ShopApiPlugin\CommandProvider;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Validator\ConstraintViolationListInterface;
+
+interface CommandProviderInterface
+{
+    public function validate(Request $request): ConstraintViolationListInterface;
+
+    public function getCommand(Request $request): object;
```
CSchulz: Even if there is no real advantage for the moment, I would like to see here some interface or abstract class instead of object. 😄

mamazu: Well, the thing might be poorly named but it is a Request to Command transformer. So it can only process webrequests. For everything else you can create a command directly.

lchrusciel: I bet @CSchulz meant some interface for commands. It would be just a markdown interface probably. The main benefit would be with type hinting here, while the drawback is that developers would have to remember about it.

mamazu: Ah yeah I will do that and replace those.

CSchulz: Yes I meant the interface for commands.

CSchulz:
Will be here more changes for the other actions?

mamazu:
@lchrusciel has to review it and comment on it.

CSchulz:
Alright, if there is nothing to do, I could take care of replacing all static command creations.

lchrusciel:
Was implemented in #475

PR #382, opened 2019-01-28
[POC] New command providers (but a bit different than those below)

pamil:
Related to discussion in https://github.com/Sylius/ShopApiPlugin/pull/375#discussion_r250996350 and #380.

This is a POC on top of #380, it removed the `validate` method from the command provider and replaces it with an exception.

In the #380 every request is validated twice: for the first time when trying to return a friendly response from a controller and for the second time when creating a command object. This might hit the performance hard, especially when using the validators which gathers data from the database (like `CartExists`). Also, we need to be sure to call `validate` every time before calling `getCommand`. 

Replacing it with an exception solves these problems and it doesn't seem to have cons.

### DIFF
--- a/src/CommandProvider/CommandProviderInterface.php
+++ b/src/CommandProvider/CommandProviderInterface.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\CommandProvider;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Validator\ConstraintViolationListInterface;
+
+interface CommandProviderInterface
+{
+    public function validate(Request $request): ConstraintViolationListInterface;
+
+    public function getCommand(Request $request): object;
+}
--- a/src/CommandProvider/CommandProviderInterface.php
+++ b/src/CommandProvider/CommandProviderInterface.php
@@ -5,11 +5,11 @@ declare(strict_types=1);
 namespace Sylius\ShopApiPlugin\CommandProvider;
 
 use Symfony\Component\HttpFoundation\Request;
-use Symfony\Component\Validator\ConstraintViolationListInterface;
 
 interface CommandProviderInterface
 {
-    public function validate(Request $request): ConstraintViolationListInterface;
-
-    public function getCommand(Request $request): object;
+    /**
+     * @throws ValidationFailedExceptionInterface
+     */
+    public function provide(Request $request): object;
 }
--- a/src/CommandProvider/DropCartCommandProvider.php
+++ b/src/CommandProvider/DropCartCommandProvider.php
@@ -8,7 +8,6 @@ use Sylius\ShopApiPlugin\Command\DropCart;
 use Sylius\ShopApiPlugin\Validator\Constraints\CartExists;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Validator\Constraints as Assert;
-use Symfony\Component\Validator\ConstraintViolationListInterface;
 use Symfony\Component\Validator\Validator\ValidatorInterface;
 
 final class DropCartCommandProvider implements CommandProviderInterface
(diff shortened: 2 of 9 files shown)


lchrusciel:
I prefer the style with two methods, to be precise what do we expect from them explicitly. I have a feeling that basing on exception is like based on the side effect. 

However, your proposal still helps a lot. Maybe it is better indeed, and I'm just defending my solution? Why way, or another I think we have something interesting, which will improve ShopApi DX. 

@mamazu, @healerz, @stefandoorn could you share your thoughts which solution do you prefer?

lchrusciel:
ping @AnderRV as well

stefandoorn:
Although I'm not so knowledged about the command structure here (only extended the Shop API for a single reason and disabled all default endpoints), but I think the provider pattern looks neat to me. Not having to run a full request twice & not being able to 'forget' calling the `validate` method sound like proper reasons to me. I'm not sure what's the alternative to working with exceptions, besides returning something like a 'Runner' object that holds validation result, validation errors & the command to be executed. The controller can then take either the validation result from there (and parse errors) or call the executor inside the 'runner'. So the provider simply returns the validation result & in case there is no error, it also returns a command to be ran. Might be similarly flexible, still not being able to forget 'validation' (you simply cannot run without validation)..

healerz:
> I prefer the style with two methods, to be precise what do we expect from them explicitly. I have a feeling that basing on exception is like based on the side effect.
> 
> However, your proposal still helps a lot. Maybe it is better indeed, and I'm just defending my solution? Why way, or another I think we have something interesting, which will improve ShopApi DX.
> 
> @mamazu, @healerz, @stefandoorn could you share your thoughts which solution do you prefer?

I prefer the style with the two methods, but without re-validating the data in the getCommand method.

Personally, i think the validation should not be part of the command provider. I think it should be possible, to just have one generic controller. In that case we don't have to be afraid, that the validate call gets forgotten.

In my mind, the controller should do something like this (generically):
```
{
    $command = $this->commandProvider->get($request);
    $violations = $this->validator->validate($command);
    if (0 < count($violations)) {
        return $this->viewHandler->handle(View::create($this->validationErrorViewFactory->create($violations), Response::HTTP_BAD_REQUEST));
    }
    $entity = $this->bus->handle($command);
    return $this->viewHandler->handle(View::create($entity, null === $entity ? Response::HTTP_NO_CONTENT : Response::HTTP_OK));
}
```

Just a brain dump, i haven't thought it through ;)

pamil:
@healerz one thing to note here - request data is validated, not the command though. Commands could have some safety checks which throw exceptions and are not user-friendly.

As for the generic controller, there would be some extra ones, because in some cases we return some data (eg. cart contents).

mamazu:
In general I agree with @pamil's approach because assuming that a function is called before hand is bad practice and calling it like you did @lchrusciel is redundant if you want to call it explicitly. However, I would still recommend to split the validate into their own function and make it protected so that the `CommandProvider` is easily extendable.

One thing I don't really like is that now the validators are defined in code and not in config. We probably need to test that as well. But looks good for me.

pamil:
@mamazu good point with validation being extendable. It is a usual use case, how would you approach it? Would the protected `validate` method has `Request` as an argument or that internal array `$rawData`? In the second case, we'd need to provide and maintain a BC promise for the internal array, both its keys and values.

lchrusciel:
@mamazu the beauty of this solution is that request object can be still supported. With `in-class` validation, This can be beneficial for end-users, because it reduce the amount of code and classes. However But ShopApi itself should provide a better way to extend validation. Therefore we may stick to request object. 

Duplication of validation is indeed overengineering in my example. I will try to incorporate your proposals to my proposition today/tomorrow.

mamazu:
Okay, I am sure that we will find a solution that is extendable and elegant.

lchrusciel:
So just as an update, new proposal https://github.com/Sylius/ShopApiPlugin/pull/380/files#diff-ed7f555ddb832ba70d4216ed946de837R15:
```php
final class DropCartCommandProvider implements CommandProviderInterface
{
    /** @var ValidatorInterface */
    private $validator;

    public function __construct(ValidatorInterface $validator)
    {
        $this->validator = $validator;
    }

    public function validate(Request $request): ConstraintViolationListInterface
    {
        return $this->validator->validate(new DropCartRequest($request));
    }

    public function getCommand(Request $request): object
    {
        $dropCartRequest = new DropCartRequest($request);

        return new DropCart($dropCartRequest->token);
    }
}
```

The problem of command validation is partially philosophical and partially due to the protection of handlers, we shouldn't allow for null values (for example) in commands.

lchrusciel:
So just as an update, new proposal https://github.com/Sylius/ShopApiPlugin/pull/380/files#diff-ed7f555ddb832ba70d4216ed946de837R15:
```php
final class DropCartCommandProvider implements CommandProviderInterface
{
    /** @var ValidatorInterface */
    private $validator;

    public function __construct(ValidatorInterface $validator)
    {
        $this->validator = $validator;
    }

    public function validate(Request $request): ConstraintViolationListInterface
    {
        return $this->validator->validate(new DropCartRequest($request));
    }

    public function getCommand(Request $request): object
    {
        $dropCartRequest = new DropCartRequest($request);

        return new DropCart($dropCartRequest->token);
    }
}
```

The problem of command validation is partially philosophical and partially due to the protection of handlers, we shouldn't allow for null values (for example) in commands.

lchrusciel:
Closing for now, but we should pick up the conversation about validation handling

Issue #354, opened 2018-11-26
Extended sylius models


healerz:
Hi everyone

I would like to discuss how to handle extended sylius models regarding the ShopApiPlugin.

### My assumption
If I add a property to a model (e.g. the order model) i see no convenient way, in the current state of the plugin, to enhance (read + write) the api to handle the additional property. Without copying and modifying the Controller, Request, Command and Handler classes, due to the final keyword.
In my eyes this would be a pity, because then i can't get updates / fixes of the affected action(s). In an extreme case, if i have a lot of modifications in base models, i end up in copying and rewriting most of the actions provided by this plugin.
I know, if it would be the "read" part only, i could handle it through the viewHandler.

If i'm wrong with my assumption please point me in the right direction :) **In this case you can stop reading :D**

#### 1. Copy and rewiring [project code only]
- Copy the involved Controller, Request, Command and Handler classes
- Add my property to all of them
- Overwrite the service definitions
- Overwrite the route definitions

#### 2. Remove the final keyword [project and plugin code]
- Remove the final keyword for controllers, requests, commands and handlers
- Maybe add a RequestFactory to be used in the controllers
- Extend (Controller), Request and Handler classes and add my property
- Rewire everything (services + routes)

#### 3. Add interfaces [project and plugin code]
- Add interfaces to requests and commands and type hint against them
- Implement the Interfaces in my custom Request and Command classes (probably decorate the existing ones, although it seems a bit pointless for these data holder classes unless the decorating part is done in a generic way)
- Create a RequestFactory to be used in the controllers
- Either replace the existing handler or decorate it
- Rewire everything

### "Conclusion"
Personally i would prefer option 2.
I would rather risk that my custom implementation breaks with updates / fixes than copying "everything" and not receiving updates and fixes.

Option 3 seems, in my eyes, a bit over engineered and depending on the exact implementation does not necessarily include updates and fixes. And again depending on the implementation, it does not prevent that something breaks with updates.

I'm looking forward to any kind of feedback! If you have other ideas / suggestions / solutions to my problem, let me know.
In case we come to a solution that requires changes in the plugin, i would be willing to create a poc :)

mamazu:
I'd suggest to use composition. So if you want to add a property to the request just wrap your own request around it.

lchrusciel:
Hello fellows, 
Sorry for such a late response. First of all, thanks for using ShopApi in the first place. I hope we will make it even better together. 

[...]

PR #384, opened 2019-01-29
Guide on how to properly extend the ShopApi

mamazu:
As the question has popped up how to extend the ShopApi this is a short guide on possible vectors on how to extend it.

### DIFF
--- a/doc/ShopApiExtensions.md
+++ b/doc/ShopApiExtensions.md
@@ -1,7 +1,7 @@
 # Extending the API
 If you want to extend the API functionality, the methodology to use depends on the part that needs to be changed.
 
-## Extending an existing view
+## Extending the factories
 If you don't want to change the underlying logic of the ShopApi and only want to extend the view by some properties, then the following steps can be taken:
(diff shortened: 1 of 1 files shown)


lchrusciel:
It is better than nothing, but I'm not sure about merging it as issues like #354 + #382 and so on will change a lot. I will keep it open so that people will have a reference to it.

mamazu:
Okay, I know that the Api is currently under construction but surely we can finalize it when the concepts are more solid.

shql:
Thanks for hinting me in the right direction here (was looking on how to adapt the shop api to our needs). As far as I can see there's a small problem with the decorated service: You are getting the cart view from the base factory and are updating this with the new property from the cart parameter. :-)

mamazu:
What issue are you talking about exactly? I have some typos in the documentation with the parameter.

shql:
Thanks! :-)

doc/ShopApiExtensions.md
```diff
@@ -0,0 +1,61 @@
...
+
+        return $cart;
+    }
+}
+```
+
+If you have customized the entity that should be used in the factory, for example in this case the `OrderItem` then you can do something like the following:
+
+```php
+    public function create(OrderItem $cart, string $locale): CartSummaryView
```
GSadee: `$cart` -> `$item` or `$orderItem`

PR #385, opened 2019-01-31
Upgrade to Sylius 1.3 & 1.4 and add support for PHP 7.3

Zales0123:
Replaces https://github.com/Sylius/ShopApiPlugin/pull/372

Some tests were failing due to a bug in Sylius with browsing products in taxon (it was already fixed, so I suppose my overridden repository would not be necessary very soon). I've fixed **almost** all tests, but there are still some issues in which products don't have all their taxons listed in a response :/ If anybody has any idea what can cause these problems, I would be happy to fix it 😄 🐎


lchrusciel:
Thanks, Mateusz! :1st_place_medal:

Zales0123:
And Bartosz and Kamil! 🎉

PR #386, opened 2019-01-31
Fixed shop api /verify-account route

nevoeiro:
In order to complete the activation of new registered users I propose to change the method of `sylius_shop_api_user_verification` route to [GET] so we are able to access the activation url sent in ShopApiPlugin:Email:verification.html.twig template, when a user registers through shop api.

### DIFF
--- a/src/Request/VerifyAccountRequest.php
+++ b/src/Request/VerifyAccountRequest.php
@@ -14,7 +14,7 @@ class VerifyAccountRequest
 
     public function __construct(Request $request)
     {
-        $this->token = $request->request->get('token');
+        $this->token = $request->get('token');
     }
 
     public function getCommand(): VerifyAccount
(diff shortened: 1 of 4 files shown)


src/Request/VerifyAccountRequest.php
```diff
@@ -14,7 +14,7 @@ class VerifyAccountRequest
 
     public function __construct(Request $request)
     {
-        $this->token = $request->request->get('token');
+        $this->token = $request->get('token');
```
mamazu: This is inefficient. If you know where the data comes from (in this case the request body) you don't need to search for it everywhere.

nevoeiro: The `token` comes from a query parameter. With this approach we support both query or request parameters. The link that is generated is like `http://shop/shop-api/US_WEB/verify-account?token=TLpPvSZOQ6GFVrkU`. So to have it working should be like `$request->query->get('token')`

mamazu: Yes, that is what I meant.

lchrusciel:
Thanks, @nevoeiro! :1st_place_medal:

PR #387, opened 2019-01-31
Moved classes into their own namespace

mamazu:
This time I refactored all the view factories.


lchrusciel:

PR #388, opened 2019-02-04
Adding validation to the `AddProductReview` endpoint

### LINKED ISSUE #160: [Bug] Review rate validation
Any value can be pass to AddReview command, therefore any value will be accepted by AddReviewHandler

mamazu:
In Sylius product reviews have to have a rating between 0 and 5. Anything else is out of range. I added validation. This should also close #160.


mamazu:
The reason why this fails is because the master also fails. Probably the dependencies have changes the way taxons are handled. We might want to consider running the unit tests every day to track those changes. Or what do you think @lchrusciel?

Zales0123:
This is the same reason as in https://github.com/Sylius/ShopApiPlugin/pull/385, there must have been some changes in Symfony or in serializer :/ I don't know, to be honest, how to fix it, although I've spent some time debugging it cc @pamil @lchrusciel 🚁

tests/Controller/AddProductReviewBySlugTest.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\Sylius\ShopApiPlugin\Controller\Product;
+
+use Symfony\Component\HttpFoundation\Response;
+use Tests\Sylius\ShopApiPlugin\Controller\JsonApiTestCase;
+
+final class AddProductReviewBySlugTest extends JsonApiTestCase
```
lchrusciel: it should be a part of ProductAddReviewBySlugApiTest

tests/Controller/AddProductReviewByCodeTest.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\Sylius\ShopApiPlugin\Controller\Product;
+
+use Symfony\Component\HttpFoundation\Response;
+use Tests\Sylius\ShopApiPlugin\Controller\JsonApiTestCase;
+
+final class AddProductReviewByCodeTest extends JsonApiTestCase
```
lchrusciel: should be part of ProductAddReviewByCodeApiTest

mamazu:
Fixed

tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -63,7 +63,7 @@ public function it_has_view_classes(): void
                 'page_links' => View\PageLinksView::class,
                 'payment' => View\PaymentView::class,
                 'payment_method' => View\PaymentMethodView::class,
-		'placed_order' => View\PlacedOrderView::class,
+        'placed_order' => View\PlacedOrderView::class,
```
lchrusciel: something still went wrong with formatting

mamazu: Fixed, strangely this wasn't caught by ECS.

lchrusciel:

PR #390, opened 2019-02-08
Add missing customer attributes to sylius_shop_api_me

nevoeiro:
In order to be able to load all user details through shopApi I suggest to change the response of `sylius_shop_api_me`, adding the missing customer attributes:  `birthday`, `gender`, `phoneNumber` and `subscribedToNewsletter`

Having this, we can get and edit all customer attributes and that will be useful if you want to use Sylius as headless system, e.g.

### DIFF
--- a/doc/swagger.yml
+++ b/doc/swagger.yml
@@ -1547,6 +1547,19 @@ definitions:
             email:
                 type: "string"
                 example: "sherlock@holmes.com"
+            birthday:
+                type: "string"
+                example: "2017-08-12"
+            gender:
+                type: "string"
+                example: "m"
+            phoneNumber:
+                type: "string"
+                example: "+490000000000"
+            subscribedToNewsletter:
+                type: "boolean"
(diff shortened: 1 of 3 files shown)


doc/swagger.yml
```diff
@@ -1547,6 +1547,19 @@ definitions:
...
+                type: "string"
+                example: "2017-08-12"
+            gender:
+                type: "string"
+                example: "m"
+            phoneNumber:
+                type: "string"
+                example: "+490000000000"
+            subscribedToNewsletter:
+                type: "integer"
```
lchrusciel: Don't we have a boolean type as well? According to tests it is not an integer https://github.com/Sylius/ShopApiPlugin/pull/390/files#diff-6d6acc52ab3732d65368f2c5d9344287R7

lchrusciel:
Thanks, @nevoeiro! :1st_place_medal:

PR #395, opened 2019-02-13
Splitting up the PHPUnit tests

mamazu:
What is this the pull request about:
* Splitting the unit tests up in the same folders as the controllers they test
* Adding return types to the test
* Adding a JSONTestCase constant to centralize the "content type header" and the "accepts" header


tests/Controller/JsonApiTestCase.php
```diff
@@ -6,6 +6,15 @@
 
 abstract class JsonApiTestCase extends \Lakion\ApiTestCase\JsonApiTestCase
```
lchrusciel: We should add `@internal` and some information to not extend this class in end projects

lchrusciel:

PR #365, opened 2019-01-03
Secure checkout complete action

lchrusciel:
This PR contains the improved implementation of #351. Current implementation will support following scenarios:

1. User is logged in - valid
2. User is not logged in and customer exists, but doesn't have an account - valid
3. User is logged in and customer exists and they are related - valid
4. User is not logged in and customer exists, but has an account - exception, he needs to log in
5. User is logged in but different mail is provided - exception

PR #400, opened 2019-02-15
Add optional auth to checkoutComplete operation

dlobato:


mamazu:
The checkout behaves different in those cases. If you are a logged in user you should not provide an email address in the request.

dlobato:
Yes, I noticed. This only says, you can use no auth or bearerAuth. Not sure this is 100% correct on openAPIv2 though, I only found the subject mentioned on openAPIv3: https://stackoverflow.com/questions/47659324/how-to-specify-an-endpoints-authorization-is-optional-in-openapi-v3

Also, maybe email should not be required because when you are logged in, the endpoint controller fill email with the logged user email.

mamazu:
Can you add a second definition with the same route but different parameters and different authentication?

dlobato:
Sure it's possible. So you suggest to have to different operations, one to complete checkout as guest (checkoutCompleteAsGuest) and one to complete checkout as logged in user (checkoutCompleteAsLoggedUser), right?

mamazu:
Well the current implementation has one route for checkout but in the handler it will throw an exception if the user is logged in and the form provides a different email (as discussed in this ticket: https://github.com/Sylius/ShopApiPlugin/pull/365#issue-242013792) The current implementation might not make it very clear.

But your pull request is completely right. This route can be accessed with no authentication and an email => guest checkout or with a logged in user and no email => "customer" checkout.

mamazu:
Thank you, David! :1st_place_medal:

Issue #264, opened 2018-03-20
Plugin does not work with disabled shop API


bitbager:
This plugin has a strong dependency on Sylius shop API in config. If someone wants to disable the default Sylius Shop API by removing the bundle dependency from Kernel, he will get an error about unsupported `sylius_shop` config.

lchrusciel:
Can you provide more details on this problem? I have tested this plugin with shop turned off, and it seems to work fine. I will try to cover this with tests in https://github.com/Sylius/SyliusShopApiPlugin/pull/271

lchrusciel:
Now, it will break the shop for sure. We need to resolve this problem: https://github.com/Sylius/ShopApiPlugin/blob/master/src/Resources/config/services.xml#L67

tophsic:
I use this bundle too with shop turned off. Without any problem… until now.

mamazu:
What is the problem now?

tophsic:
My english so bad :smile: I did not find any problem yet!

mamazu:
Well that is great. :) This issue is still relevant. I don't know how important it is to maintain compatibility with the Shop integration of Sylius.

tophsic:
Well I was wrong. I had trouble removing `ShopBundle` in my project `bundles.php` file. I did not have error mentioned in description. I missed `sylius.context.cart.session_and_channel_based` in`CompositeCartContext`. The effect was that all my orders prices were not recalculated.

I think issue title is confusing. I would rename it in "ShopApiBundle does not work if ShopBundle is disabled"

mamazu:
Well this issue should remind us to remove the ShopBundle from the dependencies of this project in order to ensure that this is really a headless plugin.

lchrusciel:
Ok, this one will be important. ShopBundle is not usable with current ShopApiPlugin implementation, due to CartContext and also a problem with recalculation. Therefore, we shouldn't require it in bundles. 

Conflicting ShopApiPlugin implementation with default is ok, but not when we still need ShopBundle in bundles list.

mamazu:
We should put that into the `README.md` so that everyone who downloads it can see it.

lchrusciel:
Mentioned in #492

PR #401, opened 2019-02-16
Double cart fix

### LINKED ISSUE #393: POST /cart with jwt create 2 order records
I try to understand how to use API, especially with cart and order process. I am a newbie on Sylius and ShopApiPlugin.

With `"sylius/shop-api-plugin": "dev-master"` and `"sylius/sylius": "~1.4.0"`, I got a failing test with following code:

```php
<?php

declare(strict_types=1);

namespace Tests\Sylius\ShopApiPlugin\Controller;

use League\Tactician\CommandBus;
use Sylius\Component\Core\Model\OrderInterface;
use Sylius\ShopApiPlugin\Command\AddressOrder;
use Sylius\ShopApiPlugin\Command\ChoosePaymentMethod;
use Sylius\ShopApiPlugin\Command\ChooseShippingMethod;
use Sylius\ShopApiPlugin\Command\CompleteOrder;
use Sylius\ShopApiPlugin\Command\PickupCart;
use Sylius\ShopApiPlugin\Command\PutSimpleItemToCart;
use Sylius\ShopApiPlugin\Model\Address;
use Symfony\Component\HttpFoundation\Response;
use Tests\Sylius\ShopApiPlugin\Controller\Utils\ShopUserLoginTrait;

final class MyTest extends JsonApiTestCase
{
    use ShopUserLoginTrait;

    private static $acceptAndContentTypeHeader = ['CONTENT_TYPE' => 'application/json', 'ACCEPT' => 'application/json'];

    /**
     * @test
     */
    public function it_creates_new_cart_linked_to_user()
    {
        $this->loadFixturesFromFiles(['shop.yml', 'customer.yml']);

        $this->logInUser('oliver@queen.com', '123password');


        $this->client->request('GET', '/shop-api/WEB_GB/orders', [], [], static::$acceptAndContentTypeHeader);
        $response = $this->client->getResponse();
        $content = json_decode($response->getContent());

        $this->assertSame(count($content), 0);

        $this->client->request('POST', '/shop-api/WEB_GB/carts', [], [], static::$acceptAndContentTypeHeader);
        $response = $this->client->getResponse();

        $this->client->request('GET', '/shop-api/WEB_GB/orders', [], [], static::$acceptAndContentTypeHeader);
        $response = $this->client->getResponse();
        $content = json_decode($response->getContent());

        $this->assertSame(count($content), 1);
    }
}
```

```bash
$ vendor/bin/phpunit -c phpunit.xml tests/Controller/MyTest.php 
PHPUnit 6.5.14 by Sebastian Bergmann and contributors.

F                                                                   1 / 1 (100%)

Time: 2.5 seconds, Memory: 30.00MB

There was 1 failure:

1) Tests\Sylius\ShopApiPlugin\Controller\MyTest::it_creates_new_cart_linked_to_user
Failed asserting that 1 is identical to 0.

/app/tests/Controller/MyTest.php:48

FAILURES!
Tests: 1, Assertions: 3, Failures: 1.
```

Observing what happened in database, I got 2 order records:

```
mysql> select * from sylius_order;
+-----+---------------------+--------------------+------------+---------------------+-------------+--------+-------+-------+-----------------------+-------------+-------------------+-------+---------------------+---------------------+---------------+-------------+----------------+---------------+----------------+--------------------------------------+-------------+
| id  | shipping_address_id | billing_address_id | channel_id | promotion_coupon_id | customer_id | number | notes | state | checkout_completed_at | items_total | adjustments_total | total | created_at          | updated_at          | currency_code | locale_code | checkout_state | payment_state | shipping_state | token_value                          | customer_ip |
+-----+---------------------+--------------------+------------+---------------------+-------------+--------+-------+-------+-----------------------+-------------+-------------------+-------+---------------------+---------------------+---------------+-------------+----------------+---------------+----------------+--------------------------------------+-------------+
| 177 |                NULL |               NULL |        481 |                NULL |         126 | NULL   | NULL  | cart  | NULL                  |           0 |                 0 |     0 | 2019-02-12 20:21:47 | 2019-02-12 20:21:47 | GBP           | en_GB       | cart           | cart          | cart           | NULL                                 | NULL        |
| 178 |                NULL |               NULL |        481 |                NULL |        NULL | NULL   | NULL  | cart  | NULL                  |           0 |                 0 |     0 | 2019-02-12 20:21:47 | 2019-02-12 20:21:47 | GBP           | en_GB       | cart           | cart          | cart           | 5758b627-ce72-4e06-b458-2597c0986d62 | NULL        |
+-----+---------------------+--------------------+------------+---------------------+-------------+--------+-------+-------+-----------------------+-------------+-------------------+-------+---------------------+---------------------+---------------+-------------+----------------+---------------+----------------+--------------------------------------+-------------+
2 rows in set (0.01 sec)
```

Maybe I misused ShopApiPlugin but it is not clear for me how I create a cart and associate it with logged user.

Thank you for your help.

mamazu:
This fixes #393 

In this pull request I introduced a new Request and Command object for picking up a cart of a customer that is logged in. This way I registered a new handler. The `PickupLoggedInCart` Command and Request object inherit from the "non logged in"-version.


mamazu:
@lchrusciel This is the current best fix. What do you think? The other approach would be to use a different context that does't automatically create a cart or exclude the cart pickup form those routes. Which I both find not optimal.

lchrusciel:
If I understood you correctly, the main issue is with `UserCartRecalculationListener`. This service is called each time we send authorization token, so it makes absolute sense. However, in this case, this solution seems to be more like a workaround. We should resolve this issue by customizing `UserCartRecalculationListener` instead. IMHO we should just turn off `UserCartRecalculationListener` for Shop API context. Still, I don't have an idea how :D 

PS. Sorry, for misclick and closing PR :)

mamazu:
Correct, the `UserCartRecalculationListener` also has a fail save, when there is no cart then there is nothing to calculate and therefore returns. The problem is that the `cartContext` that the listener is provided with, doesn't throw an exception if the cart is not found but silently creates a new one.

mamazu:
Alternatively you could also remove the `sylius.context.cart.new` context from the composite cart (see context.https://github.com/Sylius/Sylius/issues/10192) which is the best solution in my opinion.

lchrusciel:

tophsic:
Big thanks!

src/Resources/config/services.xml
```diff
@@ -63,5 +63,10 @@
             <argument type="service" id="validator" />
             <argument type="service" id="tactician.commandbus" />
         </service>
+
+        <!-- Removing the create cart context from composite context (see: https://github.com/Sylius/Sylius/issues/10192) -->
+        <service id="sylius.context.cart.new" class="Sylius\Component\Order\Context\CartContext">
```
tophsic: What happens if a project use in the same time shop bundle and shop api? This create cart context is removed for both context, isnt'it?

This how I understand your comment https://github.com/Sylius/Sylius/issues/10192#issuecomment-467337181

mamazu: If you want to use the shop front-end and the API as well those will collide. Or so we think. I have not tested it.

mamazu: But we also have a ticket for removing the shop from the API package as a dependency (#264)

lchrusciel: @tophsic that's correct, right now ShopApi will collide with default shop implementation

Issue #190, opened 2017-07-18
Change bundle's configuration key


pamil:
It's `shop_api` now, but it should be prepended with vendor to be `sylius_shop_api`.

mamazu:
This will be a bc break. But I guess this should be fine if we mention it in the change logs.

mamazu:
Done.

Issue #340, opened 2018-11-08
Resurrect UPGRADE.md


lchrusciel:
We need to document changes such as https://github.com/Sylius/ShopApiPlugin/pull/339

mamazu:
An `UPGRADE.md` already exists but I don't know which version it upgrades to because we don't have a tagged version.

lchrusciel:
My fault. We need to resurrect this file then.

lchrusciel:
I need to process this problem and figure it out. Not sure what will be the best option right now

mamazu:
I think for now that we should just list all the changes and then we can still add Headings for the versions later on. What do you think?

mamazu:
I think this issue can be closes as we are now taking care of the read me one merge request at a time. :D

PR #404, opened 2019-03-13
Changing the config key as suggested by pamil #190

mamazu:
THIS IS A BC BREAK and needs to be documented. (How?)

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,6 @@
+# UPGRADE FROM 1.0.0-beta.18 to 1.0.0-beta.20
+* The configuration key for the shop api is now `sylius_shop_api`.
+
 # UPGRADE FROM 1.0.0-beta.17 to 1.0.0-beta.18
 
 * Customer registration payload changed:
--- a/README.md
+++ b/README.md
@@ -149,9 +149,9 @@ The latest documentation is available [here](https://app.swaggerhub.com/apis/Syl
 
 ### Attributes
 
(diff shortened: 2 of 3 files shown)


lchrusciel:
Fixing #263, probably we should document some UPGRADE.md file

lchrusciel:
Also, it should be adjusted at least [here](https://github.com/Sylius/ShopApiPlugin#attributes)

mamazu:
> Fixing #263, probably we should document some UPGRADE.md file

But for which version?

lchrusciel:
We need to create a new record in https://github.com/Sylius/ShopApiPlugin/blob/master/UPGRADE.md, and notice all the changes, that are BC-Breaks. Problem is that we are over 200 commits from the last release. Nevertheless, some changes that are pretty straightforward to describe. 

(just to keep a track #340)

mamazu:
Can you generate a list of all the commit messages and I will try to craft a `UPGRADE.md` that reflects those changes.

UPGRADE.md
```diff
@@ -1,3 +1,6 @@
+# UPGRADE FROM 1.0.0-beta.18 to 1.0.0-beta.20
```
lchrusciel:
```suggestion
# UPGRADE FROM 1.0.0-beta.21 to 1.0.0-beta.22
```

lchrusciel:
Thank you, @mamazu! :tada:

PR #407, opened 2019-03-14
[WIP] Fix customer group rule promotion

tophsic:
If we got a promotion with a customer group rule, we have to know cart is linked to a customer to apply this promotion. If user is logged in, cart should know from time it is created.


tophsic:
Actually, customer is linked to a cart only on `PickupCartHandler`. How can I factorize with other relevent handlers, such `PutSimpleItemToCartHandler`?

tophsic:
@mamazu `CartPickupApiTest::it_creates_one_cart_linked_to_user_if_user_is_logged_in` fail without changes I made on `PickupCartHandler`. Do you think it is how we can fix this? I have to do the same thing in other cart handlers but want to find a solution to do not duplicate code.

mamazu:
Well you can also checkout without being logged in. So we also still need the pickup endpoint. But as @tophsic has correctly pull-requested. If a customer is logged in it should be associated to the cart. But on the other hand if the customer already has a cart and then logs in it also has to associate the cart with the logged in customer.

mamazu:
Should be solved with #488

PR #414, opened 2019-03-26
Adding coupon code to cart view

mamazu:
Adding the applied coupon code to the cart view.


lchrusciel:
Thanks, @mamazu! :tada:

PR #419, opened 2019-04-04
[Maintenance] Switch to Symfony Messenger

### LINKED ISSUE #402: Replace Tactician with Symfony Messenger

GSadee:
Fixes #402

### DIFF
--- a/src/Handler/Cart/AddCouponHandler.php
+++ b/src/Handler/Cart/AddCouponHandler.php
@@ -40,7 +40,7 @@ final class AddCouponHandler
     }
 
     /** @param AddCoupon $addCoupon */
-    public function handle(AddCoupon $addCoupon)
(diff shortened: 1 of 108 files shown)


src/Handler/Cart/AddCouponHandler.php
```diff
@@ -40,7 +40,7 @@ public function __construct(
     }
 
     /** @param AddCoupon $addCoupon */
-    public function handle(AddCoupon $addCoupon)
+    public function __invoke(AddCoupon $addCoupon): void
```
lchrusciel: We may also include this change in UPGRADE, same with `handle` -> `dispatch` change

GSadee: I've updated UPGRADE file

lchrusciel:
Thanks, Grzegorz! :1st_place_medal:

PR #423, opened 2019-04-09
[Maintenance] Enable some tests

GSadee:


src/Controller/Checkout/ShowAvailablePaymentMethodsAction.php
```diff
@@ -29,25 +32,33 @@ final class ShowAvailablePaymentMethodsAction
...
     }
 
     public function __invoke(Request $request): Response
     {
         /** @var OrderInterface $cart */
         $cart = $this->cartRepository->findOneBy(['tokenValue' => $request->attributes->get('token')]);
 
-        $payments = [];
+        if (!$this->isCheckoutTransitionPossible($cart, OrderCheckoutTransitions::TRANSITION_SELECT_PAYMENT)) {
+            throw new BadRequestHttpException('The payment methods cannot be resolved in the current state of cart!');
```
mamazu: I would suggest to use a different exception to not tie this logic to the HTTP - Layer. Isn't there a general transition not possible exception?

lchrusciel: We are in the controller, so coupling with HTTP is totally ok. We could, however, add a more generic exception like `Transition not possible`, but this one is fair enough for me.

mamazu: Ah okay, didn't notice that.

lchrusciel:
Thank you, Grzegorz! :tada:

PR #425, opened 2019-04-09
Change parameter name to make it consistent with reset password

dlobato:
The body parameter is always called content on all endpoints.


lchrusciel:
Thanks, David! :1st_place_medal:

PR #427, opened 2019-04-11
[Routing] Remove deprecated product and cart routes

GSadee:


src/Request/Cart/PickupCartRequest.php
```diff
@@ -18,10 +18,6 @@ class PickupCartRequest
 
     public function __construct(Request $request)
     {
-        if ($request->attributes->has('token')) {
```
mamazu: You can probably also remove the `RequestCartTokenNormalizer` service

GSadee: I'm not sure, because `RequestCartTokenNormalizer` service is used in `PutItemToCartAction` and `PutItemsToCartAction` actions and it is responsible for dispatching `PickupCart` command if the token is not provided. So I guess that this behaviour is still useful if the cart has not been picked up before 😃

mamazu: Okay.

tests/Controller/Cart/CartEstimateShippingTest.php
```diff
@@ -27,11 +27,10 @@ public function it_returns_not_found_exception_if_cart_has_not_been_found(): voi
      */
     public function it_calculates_estimated_shipping_cost_based_on_country(): void
     {
-        $this->loadFixturesFromFiles(['shop.yml', 'country.yml', 'shipping.yml']);
+        $this->loadFixturesFromFiles(['shop.yml', 'cart.yml', 'country.yml', 'shipping.yml']);
 
-        $token = 'SDAOSLEFNWU35H3QLI5325';
+        $token = 'PICKEDUPCARTTOKEN';
 
-        $this->pickupCart($token);
```
lchrusciel: Shouldn't we pick up a cart with API, and use provided token to make rest of calls?

GSadee: I've added dispatching `PickupCart` and `PutSimpleItemToCart`

lchrusciel:
Thank you, Grzegorz! :1st_place_medal:

PR #429, opened 2019-04-17
Add .editorconfig

teohhanhui:


mamazu:
I don't think that we need this. We have the Easy Coding Standard Package (ECS) with a list of code style guidelines that will also be checked in the CI run.

teohhanhui:
EditorConfig is supported by most (if not all) IDEs and text editors. It's not used to enforce coding style or standards, but it takes care of correct indentation etc.

If it'd help convince you, it's already in Sylius and Sylius-Standard.

lchrusciel:
Thank you, Teoh! :tada:

PR #437, opened 2019-04-30
[Maintanance] Remove redundant specs

lchrusciel:
TypeError tests don't provide any value, we are just testing that PHP itself is working correctly


GSadee:
Thanks, Łukasz! :tada:

PR #439, opened 2019-04-30
Cart summary now only works for orders in state cart

### LINKED ISSUE #438: cartSummarize allows access to all orders
cartSummarize endpoint (/{channelCode}/carts/{token} get) allows to access any order regardless of state. It should only allow to access orders in cart state.

[cartSummarize_response.json.zip](https://github.com/Sylius/ShopApiPlugin/files/3132149/cartSummarize_response.json.zip) is a response to a random order created by fixtures data to the endpoint:
GET http://127.0.0.1:8000/shop-api/US_WEB/carts/pcBfQSkq8t

mamazu:
- Also added local phpunit config to the `.gitignore`
Closes #438


mamazu:
@lchrusciel The repository now uses the query builder which can not be tested but I think that getting a cart from the order repository should be inside the Sylius Core. What do you think?

src/ViewRepository/Cart/CartViewRepository.php
```diff
@@ -29,7 +30,14 @@ public function __construct(
     public function getOneByToken(string $orderToken): CartSummaryView
     {
         /** @var OrderInterface $cart */
-        $cart = $this->cartRepository->findOneBy(['tokenValue' => $orderToken]);
+        $cart = $this->cartRepository->createCartQueryBuilder()
+            ->andWhere('o.checkoutState != :checkoutState')
+            ->andWhere('o.tokenValue = :tokenValue')
+            ->setParameter('tokenValue', $orderToken)
+            ->setParameter('checkoutState', OrderCheckoutStates::STATE_COMPLETED)
```
GSadee: `->setParameter('state', OrderInterface::STATE_CART)`?

src/ViewRepository/Cart/CartViewRepository.php
```diff
@@ -29,7 +30,14 @@ public function __construct(
     public function getOneByToken(string $orderToken): CartSummaryView
     {
         /** @var OrderInterface $cart */
-        $cart = $this->cartRepository->findOneBy(['tokenValue' => $orderToken]);
+        $cart = $this->cartRepository->createCartQueryBuilder()
+            ->andWhere('o.checkoutState != :checkoutState')
```
GSadee: `->andWhere('o.state = :state')`?

src/ViewRepository/Cart/CartViewRepository.php
```diff
@@ -29,7 +30,14 @@ public function __construct(
     public function getOneByToken(string $orderToken): CartSummaryView
```
GSadee: `$orderToken` -> `$token`

tests/Controller/Cart/CartSummarizeApiTest.php
```diff
@@ -112,7 +132,7 @@ public function it_shows_summary_of_a_cart_filled_with_a_product_with_variant():
         $bus->dispatch(new PickupCart($token, 'WEB_GB'));
 
         $variantWithOptions =
-<<<EOT
+            <<<EOT
```
GSadee: Unnecessary change

src/ViewRepository/Cart/CartViewRepository.php
mamazu: For some reason this does not work. It still shows also complete orders.

tests/Controller/Cart/CartSummarizeApiTest.php
```diff
@@ -154,7 +174,7 @@ public function it_shows_summary_of_a_cart_filled_with_a_product_with_variant_in
         $bus->dispatch(new PickupCart($token, 'WEB_DE'));
 
         $variantWithOptions =
-<<<EOT
+            <<<EOT
```
GSadee: Unnecessary change

src/ViewRepository/Cart/CartViewRepository.php
GSadee: Do you mean tests? Order with `ORDERNUMBERPLACED` number has `completed` checkoutState, but it has `cart` as a state what is a mistake I guess.

IMO you should check `state` instead of `checkoutState` field, which should work correctly.

mamazu: Yes, this makes sense so I have to fix the fixtures that define the test?

GSadee: I guess yes, because the state of this order is not correct imo

src/ViewRepository/Cart/CartViewRepository.php
mamazu: This is handled in the `createCartQueryBuilder`.

src/ViewRepository/Cart/CartViewRepository.php
mamazu: I changed it but now the tests fail when trying to display an order. The totals are missing now. Is that normal for an order in the state of new?

src/ViewRepository/Cart/CartViewRepository.php
```diff
@@ -26,10 +27,15 @@ public function __construct(
         $this->cartViewFactory = $cartViewFactory;
     }
 
-    public function getOneByToken(string $orderToken): CartSummaryView
+    public function getOneByToken(string $token): CartSummaryView
     {
         /** @var OrderInterface $cart */
-        $cart = $this->cartRepository->findOneBy(['tokenValue' => $orderToken]);
+        $cart = $this->cartRepository->createCartQueryBuilder()
```
lchrusciel: Not sure if `$this->cartRepository->findOneBy(['tokenValue' => $token, 'state' => OrderInterface::STATE_CART]);` wouldn't be enough

mamazu: True. Then I can also add the test back in.

lchrusciel:
I will pick it up tomorrow

tests/Controller/Utils/OrderPlacerTrait.php
```diff
@@ -0,0 +1,50 @@
...
+            'postcode' => 'NWB',
+            'provinceName' => 'Greater London',
+        ])
+        ));
+        $bus->dispatch(new ChooseShippingMethod($token, 0, 'DHL'));
+        $bus->dispatch(new ChoosePaymentMethod($token, 0, 'PBC'));
+        $bus->dispatch(new CompleteOrder($token, $email));
+    }
+
+    abstract protected function get($id);
```
GSadee: Add return type

lchrusciel: Unfortunately, it is not possible :( 
```
PHP Fatal error:  Declaration of Tests\Sylius\ShopApiPlugin\Controller\JsonApiTestCase::get($id) must be compatible with Tests\Sylius\ShopApiPlugin\Controller\Utils\OrderPlacerTrait::get(string $id): object in tests/Controller/Order/OrderShowApiTest.php on line 12 
```

Should I add a comment about it?

lchrusciel:
Thanks, @mamazu! :tada:

mamazu:
Yeah, that was an obvious fix. Thanks.

PR #442, opened 2019-05-02
adding company and phoneNumber to checkout/address

hashnz:


mamazu:
Are those options now required or are they optional?

hashnz:
@mamazu they should be optional, there's a test here that doesn't pass them in https://github.com/Sylius/ShopApiPlugin/blob/762ecd752154e97d1ce42a3c2dc8a3da572c6e8f/tests/Controller/Checkout/CheckoutSummarizeApiTest.php#L112

dlobato:
new fields are not reflected on swagger definition

hashnz:
@dlobato thanks, I've added those

mamazu:
Thanks, Matt! :tada:

Issue #428, opened 2019-04-16
ShowOrderDetails requires authentication


dlobato:
We are implementing checkout as a guest user and we realized that ShowOrderDetails endpoint (/{channelCode}/orders/{tokenValue} get) requires authentication, therefore after complete as a guest there is no way to access order details. But, using the carts endpoint (/{channelCode}/carts/{token} get) we can access to the placed order information anyway (it doesn't check if order is on cart state).

The question is: is it really needed to authenticate on the order details endpoint? if so, how do you suggest to implement this for guest users?

lchrusciel:
I would say, that cart endpoint should care only about orders in cart state. 

And according to order endpoint, it should allow accessing guest orders, while checking authorization for carts placed by users with account.

mamazu:
We have also noticed this issue that a guest can not access their order view. The problem with that is if you open the endpoint for guest carts to have no authentication, then you would also be able to see the orders of other guests without any authentication and this would be a huge problem because on the orders there is sensitive information like the address.
What I would suggest for guest checkouts would be: The customer gets an order confirmation mail anyways so why not include some kind of token so that for guest orders the user has to also provide a token to view the order summary.

And for the cart... it's still under development. :D

dlobato:
The order token is already a random hash, so it is not like you can access an order by randomly trying.
We need to access the order detail right after complete, to show the order number or to allow changing payment method in case of error, so the email solution wouldn't be ideal in our use case.

mamazu:
Hm, so the currently best solution to get the order number of an order on the cart complete endpoint is just to override or decorate the endpoint and then after the order is created query the database for the most recent order. This is not the most beautiful solution but this is the way to do it as the command that creates the order can/should not return the order (or it's number).

dlobato:
Well, I understand I can override the endpoint to do whatever fits my requirements, but I'd rather use standard behavior and that's why I'm discussing the subject here.

/{channelCode}/carts/{token} get allows you to access any order (regardless of order state) information given that you have the right token. So, it seems clear that this is not the intended behavior and it should be marked as a bug (carts endpoint should check order is in cart state).

On the other hand, I still believe that /{channelCode}/orders/{tokenValue} get endpoint authentication is not needed as tokenValue can't be randomly guessed and therefore only the client that created the order have access to it. I might be wrong, could you give me a good reason to change my mind?

mamazu:
[...]

PR #443, opened 2019-05-06
ShowOrderDetailsAction handles guest customer orders.

dlobato:
Following conversation #428, this is an initial attempt to allow guest to access order details.


src/Controller/Order/ShowOrderDetailsAction.php
```diff
@@ -37,20 +37,33 @@ public function __construct(
...
-            ;
-        } catch (TokenNotFoundException $exception) {
-            return $this->viewHandler->handle(View::create(null, Response::HTTP_UNAUTHORIZED));
+        try {
+            if (null !== $user) {
+                $order = $this
+                    ->placedOrderQuery
+                    ->getOneCompletedByCustomerEmailAndToken($user->getEmail(), (string) $request->attributes->get('tokenValue'))
+                ;
+            } else {
```
lchrusciel: Can we extract this logic to a private function and use early return?

dlobato: Sure

src/Controller/Order/ShowOrderDetailsAction.php
```diff
@@ -37,20 +38,39 @@ public function __construct(
...
-                ->getOneCompletedByCustomerEmailAndToken($user->getCustomer()->getEmail(), (string) $request->attributes->get('tokenValue'))
-            ;
-        } catch (TokenNotFoundException $exception) {
-            return $this->viewHandler->handle(View::create(null, Response::HTTP_UNAUTHORIZED));
+        try {
+            $order = $this->getPlacedOrderView(
+                (string) $request->attributes->get('tokenValue'),
+                $user
+            );
         } catch (\InvalidArgumentException $exception) {
```
mamazu: I am not really sure, do we want to have a better handling here if the token belongs to another user. So a 404 for an unknown token and a 403 Unauthorized for tokens that need login maybe.

src/Controller/Order/ShowOrderDetailsAction.php
```diff
@@ -37,20 +38,39 @@ public function __construct(
 
     public function __invoke(Request $request): Response
     {
-        try {
+        $groups = [GroupsExclusionStrategy::DEFAULT_GROUP];
```
mamazu: Very nice idea to handle this logic with view groups. Didn't know that was an option. :+1:

mamazu:
Then this should be ready to be merged.

dlobato:
Yes

mamazu:
Thank you, David! :tada:

PR #446, opened 2019-06-05
Require Messenger ~4.3.0

pamil:
There are BC breaks between Messenger 4.2 and 4.3.

PR #448, opened 2019-06-06
[Checkout] Order complete fix

lchrusciel:


src/Controller/Checkout/CompleteOrderAction.php
```diff
@@ -48,13 +49,17 @@ public function __invoke(Request $request): Response
...
                 )
             );
-        } catch (WrongUserException $notLoggedInException) {
-            return $this->viewHandler->handle(
-                View::create(
-                    'You need to be logged in with the same user that wants to complete the order',
-                    Response::HTTP_UNAUTHORIZED
-                )
-            );
+        } catch (HandlerFailedException $handlerFailedException) {
```
bartoszpietrzak1994: Is it possible to turn off the messenger's exception wrapper? This solution leaves us vulnerable to changes in Symfony Messenger, which is the experimental component.

pamil: BC breaks in the Messenger component happen between minor releases and in #446 I've blocked it to 4.3.*, so we should be safe. (anyway #446 includes the same fix 🎉)

lchrusciel:
Closing in favour of #446

PR #449, opened 2019-06-08
fixed product reviews by product code route by api convention

antonioperic:


mamazu:
Thank you for your contribution and welcome to the community. Your pull request looks great.
Could you please add an entry to the `UPGRADE.md` file for the changes?  This would provide a good way to notice the route changes when upgrading.

src/Resources/config/routing/productByCode.yml
```diff
@@ -21,3 +9,15 @@ sylius_shop_api_add_product_review_by_code:
     methods: [POST]
     defaults:
         _controller: sylius.shop_api_plugin.controller.product.add_review_by_code_action
+
```
mamazu: Why was that moved?

antonioperic: /reviews is more specific route so it is moved on top, cause in some situaion routing can consider /reviews as part of slug/code
in this situation that will never happen
and also it is better/faster to have more specific routes on top

mamazu: Yeah, that makes sense.

mamazu:
Thank you, Antonio! :tada:

PR #450, opened 2019-06-08
fix product not assigned to channel error message

CSchulz:
I was wondering why querying an existing product results into a channel not found error.

(Could be possible that there will be some test failures, I will fix them soon).


mamazu:
Very helpful error message. Thank you.

mamazu:
Thanks, @CSchulz! :1st_place_medal:

CSchulz:
Perhaps we should add a test for it. 😆

mamazu:
This would also be a great pull request. That would explain why the tests didn't cover it.

Issue #433, opened 2019-04-28
Address model is not being validated at Checkout\AddressAction step


diimpp:
Controller validates [`Request\Checkout\AddressOrderRequest` ](https://github.com/Sylius/ShopApiPlugin/blob/master/src/Controller/Checkout/AddressAction.php#L44), which keeps billing/shipping address as array, therefore `Model\Address` validation rules are not applied.

If initialized `Model\Address` will be added to `AddressOrderRequest`, then those models can be validated by association by current controller without any change anywhere else.

P.S. I'm basing my finding on older plugin version, but seems to be the case with current version as well.

mamazu:
Thank you for reporting this issue. You are correct the conversion from the raw request data (array) to the Address entity is done in the `getCommand` method which is called after the validation has finished.

mamazu:
This has been done in the merge request #453

PR #453, opened 2019-06-21
Adding address mapper

### LINKED ISSUE #424: assertCountryExists
The Addressbook handlers assert that the country exists, but the checkout handlers don't do that.

Shouldn't they behave the same?

mamazu:
Currently the api has a few places where addresses are introduced into the system. In this pull request there is one service that handles it.

Closes #424

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
     "require": {
         "php": "^7.2",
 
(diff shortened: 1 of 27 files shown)


mamazu:
Closes #433

lchrusciel:
Thanks, @mamazu! :tada:

PR #454, opened 2019-06-22
Fixing double carts

mamazu:
In the last pull request #401 the cart provider was overridden however this fix is much more specific in that it replaces the cart context only in a specific point (the cart blamer) The cart blamer is used on login so it is also the point where the cart creation shouldn't occur. For a more indepth view on if this is the right part of where the override should happen, the Sylius Core team would need to give feedback. This however should prevent the Sylius frontend from crashing when using the ShopApi

### DIFF
--- a/src/DependencyInjection/Compiler/CartContextPass.php
+++ b/src/DependencyInjection/Compiler/CartContextPass.php
@@ -9,25 +9,24 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Reference;
 
-class CartContextPass implements CompilerPassInterface
+final class CartContextPass implements CompilerPassInterface
 {
     public const CART_CONTEXT_SERVICE_TAG = 'sylius.context.cart';
-
-    private $compositeId = 'sylius.shop_api_plugin.context.cart';
-    private $tagName = self::CART_CONTEXT_SERVICE_TAG;
+    private const COMPOSITE_ID = 'sylius.shop_api_plugin.context.cart';
+    private const EXCLUDED_SERVICE = 'sylius.context.cart.new';
 
     /**
      * {@inheritdoc}
      */
     public function process(ContainerBuilder $container): void
     {
-        if (!$container->has($this->compositeId)) {
+        if (!$container->has(self::COMPOSITE_ID)) {
             return;
         }
 
-        $contextDefinition = $container->findDefinition($this->compositeId);
+        $contextDefinition = $container->findDefinition(self::COMPOSITE_ID);
 
-        $taggedServices = $container->findTaggedServiceIds($this->tagName);
+        $taggedServices = $container->findTaggedServiceIds(self::CART_CONTEXT_SERVICE_TAG);
         foreach ($taggedServices as $id => $tags) {
             $this->addMethodCalls($contextDefinition, $id, $tags);
         }
@@ -35,7 +34,7 @@ class CartContextPass implements CompilerPassInterface
 
     private function addMethodCalls(Definition $contextDefinition, string $id, array $tags): void
     {
-        if ($id === 'sylius.context.cart.new') {
+        if ($id === self::EXCLUDED_SERVICE) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


src/DependencyInjection/Compiler/CartContextPass.php
```diff
@@ -0,0 +1,46 @@
...
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+
+class CartContextPass implements CompilerPassInterface
+{
+    public const CART_CONTEXT_SERVICE_TAG = 'sylius.context.cart';
+
+    private $compositeId = 'sylius.shop_api_plugin.context.cart';
```
lchrusciel: What would you say about defining it as a private const?

lchrusciel: What would you say about declaring it as a private const?

src/DependencyInjection/Compiler/CartContextPass.php
```diff
@@ -0,0 +1,46 @@
...
+    private $tagName = self::CART_CONTEXT_SERVICE_TAG;
```
lchrusciel: Do we need this variable?

src/DependencyInjection/Compiler/CartContextPass.php
```diff
@@ -0,0 +1,46 @@
...
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+
+class CartContextPass implements CompilerPassInterface
```
lchrusciel: Can we finalize it?

src/DependencyInjection/Compiler/CartContextPass.php
```diff
@@ -0,0 +1,46 @@
...
+
+        $taggedServices = $container->findTaggedServiceIds($this->tagName);
+        foreach ($taggedServices as $id => $tags) {
+            $this->addMethodCalls($contextDefinition, $id, $tags);
+        }
+    }
+
+    private function addMethodCalls(Definition $contextDefinition, string $id, array $tags): void
+    {
+        if ($id === 'sylius.context.cart.new') {
```
lchrusciel: Can be declared as const as well

src/Resources/config/services.xml
```diff
@@ -71,8 +71,13 @@
         </service>
 
         <!-- Removing the create cart context from composite context (see: https://github.com/Sylius/Sylius/issues/10192) -->
-        <service id="sylius.context.cart.new" class="Sylius\Component\Order\Context\CartContext">
-            <argument type="service" id="sylius.factory.order" />
+        <service id="sylius.shop_api_plugin.context.cart" class="Sylius\Component\Order\Context\CompositeCartContext" />
```
lchrusciel: It should work without these services. May you test it?

mamazu: Yes, we need it. The compiler pass doesn't create the service but only uses it.

lchrusciel: But the service is created by another compiler pass, shouldn't it be available?

src/Resources/config/services.xml
```diff
@@ -71,8 +71,13 @@
         </service>
 
         <!-- Removing the create cart context from composite context (see: https://github.com/Sylius/Sylius/issues/10192) -->
```
lchrusciel: can be removed

src/Resources/config/services.xml
mamazu: No compiler pass creates it. The compiler pass I wrote just uses this and adds the other context.

lchrusciel: I meant \Sylius\Bundle\OrderBundle\DependencyInjection\Compiler\RegisterCartContextsPass

mamazu: It could be that I could use this class to create the service as well. I just took the logic of the `PrioritizedCompositeServicePass` and applied it here. This requires the service to exist though.

lchrusciel:
Fixed in #490, thanks to Max for your work anyway :)

Issue #452, opened 2019-06-19
Guest authentication for order details


JakobTolkemit:
Our take on https://github.com/Sylius/ShopApiPlugin/issues/428

We could add a new firewall for guest authentication using jwt. Instead of a user it would provide a customer.

The token would be generated on order completion and included in the order confirmation email.
_In an implementation it would probably be a link to a guest order details page with the token as query parameter._

In order to get a new token (because jwt are valid for a limited time only) the customer would have to provide specific information for that order (e.g. used email address, order number & used payment method).

PR #457, opened 2019-07-04
Guest authentication with jwt

JakobTolkemit:
A guest authentication like proposed in #452 

We implemented this in our project and these are basically the parts that we thought make sense to give back to the plugin.


mamazu (approved):
Looks very good.

mamazu:
@dlobato What do you think?

dlobato:
I believe #443 already solves the problems mentioned on #428
This is a different feature, or Am I seeing this wrong?

mamazu:
This pull request is more elaborate and offers a guest customer to login as well (with a jwt token and so on). It requires the customer to provide some information about the order (and not just the order uuid) to confirm that it is indeed the guest who orders it.

dlobato:
The use case that I'm more concerned about is "changing payment after payment failure on guest checkout". So, as far as this is covered with this change I'm fine with it. I'd suggest to make authentication with guest users optional, showing only a limited view of orders as implemented in #443. This way we can keep the checkout process simple and still show the complete order to a guest user if she provides enough information to trust her.

Also, @lchrusciel stated: "According to orders endpoint, I would say, that we should stick as close as possible with the default implementation", and in my eyes this PR already moves a little bit away from it.

kortwotze:
Co-dev for this PR here: So this PR does not fiddle around with payment or anything - this is just there to enable a guest to check on his order after a period of time with protected access through a guest-firewall.
I am sure one can extend on this to enable order modification for guest orders - and in my personal opinion failed payments and a subsequent change of the payment method should be handled before an order is created (e.g. PayPal, CC and other instant payments). For other payments, an order should be cancelled and reordered instead of changing the payment, what do you think?

mamazu:
This could be closed in favor of #443 which has been merged.

kortwotze:
@lchrusciel just pinging you again to see this :slightly_smiling_face: :rocket:

JakobTolkemit:
@lchrusciel Still no update on this?

PR #460, opened 2019-07-11
More LoggedInUserProvider implementation and adding phpstan back in

mamazu:
Adding the user provider to the `CustomerDetailsAction` and adding the phpstan changes back in.


tests/Responses/Expected/customer/logged_in_customer_details_response.json
```diff
@@ -1,4 +1,5 @@
 {
+    "id": "@integer@",
```
lchrusciel: We don't need to expose `id` to logged in customer

lchrusciel:

PR #462, opened 2019-07-16
[RFC] Rework address book from commands to crud

Tomanhez:


mamazu:
Thanks for your pull request, the idea is great the problem I see though is that this changes the response codes and the response structure very drastically.

tests/Responses/Expected/address_book/validation_create_address_book_response.json
```diff
@@ -1,25 +1,40 @@
...
-    "postcode": [
-      "@string@"
-    ]
-  }
+    "children": {
+      "firstName": {
+        "errors": [
+          "Please enter first name."
+        ]
+      },
```
lchrusciel: Something went wrong with the formatting in this file. Can we clean it?

mamazu:
Well the validation message structure is different. I'd say that this is a problem indeed. If there is a way to hook into that, this would be great other than that this would cause a very big inconsistency with the current validation views, which I wouldn't really like.

src/Resources/config/app/config.yml
```diff
@@ -10,6 +10,9 @@ jms_serializer:
             sylius-shop-api-placed-order:
                 namespace_prefix: "Sylius\\ShopApiPlugin\\View\\Order"
                 path : "@ShopApiPlugin/Resources/config/serializer/Order"
+            sylius-shop-api-sylius-core:
```
lchrusciel: Should be removed

tests/Controller/AddressBook/AddressBookCreateAddressApiTest.php
```diff
@@ -38,7 +38,7 @@ public function it_allows_user_to_add_new_address_to_address_book(): void
         }
 EOT;
 
-        $this->client->request('POST', '/shop-api/WEB_GB/address-book', [], [], self::CONTENT_TYPE_HEADER, $data);
+        $this->client->request('POST', '/shop-api/WEB_GB/address-book/', [], [], self::CONTENT_TYPE_HEADER, $data);
```
lchrusciel: We should mention about this change in [UPGRADE.md](https://github.com/Sylius/ShopApiPlugin/blob/master/UPGRADE.md#upgrade-from-100-beta21-to-100-beta22)

tests/Responses/Expected/address_book/show_address_book_response.json
```diff
@@ -3,39 +3,44 @@
...
         "firstName": "Jeanie",
         "lastName": "Metz",
+        "phoneNumber": "349713",
         "countryCode": "GB",
+        "provinceCode": "GB-SCT",
         "street": "Kupreska",
         "city": "Klingside",
-        "postcode": "33553",
-        "provinceCode": "GB-SCT",
-        "company": "Locastic",
```
lchrusciel: Oh heck, we are losing some part of the information. Can you investigate it?

tests/Responses/Expected/address_book/validation_create_address_book_response.json
lchrusciel: up

tests/Responses/Expected/address_book/show_address_book_response.json
mamazu: Maybe because it is a different serializer configuration?

UPGRADE.md
```diff
@@ -40,9 +40,12 @@
     | `{channelCode}/verify-account`           | `verify-account`                    |
     | `{channelCode}/taxon-products/*`         | `taxon-products/*`                  |
     | `{channelCode}/taxons/*`                 | `taxons/*`                          |
+   
+* The address-book create route have been changed :
```
GSadee:
```suggestion
* The address-book create route has been changed :
```

tests/Controller/AddressBook/AddressBookCreateAddressApiTest.php
```diff
@@ -28,13 +28,14 @@ public function it_allows_user_to_add_new_address_to_address_book(): void
         $data =
 <<<JSON
         {
-            "firstName": "Jurica",
-            "lastName": "Separovic",
-            "phoneNumber": "091892212",
-            "city": "Split",
-            "street": "Kupreska 12",
+            "firstName": "New name",
```
GSadee: Unnecessary changes

lchrusciel:
Let's merge it and see what feedback from the community we will receive. In any case, we can always revert this change.

lchrusciel:
Thanks, Tomasz! :1st_place_medal:

diimpp:
Why this change was necessary? How serialization will handle case of different data format for SyliusAdminApi and SyliusShopApi? Now I need to start serialization groups hell again, only at ShopApi side.

Is validation consistent with rest of the ShopApi output as @lchrusciel asked before?

-1 for the change, I think I will keep original implementation.

lchrusciel:
No, validation is not consistent yet. 

Thanks a lot for your opinion. I'm not insisting on this implementation, I'm ready to move back to the previous one if this will be the request from the community. Of course, I want to make a decision about it before 1.0

diimpp:
Thank you for clarification.

Base output have changed as well as address renders customer details (and other unnecessary details through customer relationship).

To clarify my point,

Conceptually, till this change, ShopApi was "add-on" (I'm aware, that it's not really compatible with SyliusShop, but that's a technical difficulty, not intentional change), but right now it overrides parent Sylius serialization mappings for Address

```yaml
jms_serializer:
    metadata:
        directories:
            # ...
            sylius-addressing:
                namespace_prefix: "Sylius\\Component\\Addressing"
                path: "@SyliusShopApiPlugin/Resources/config/serializer/Address"
```
which makes it intentionally tamper with Sylius Core.

I think such decision for conception shift should be weighted closely and in my opinion a few boiler-plate classes around address book are not really a reason for this approach.

PR #465, opened 2019-07-17
[Order] Remove channel code from order routes

GSadee:

PR #469, opened 2019-07-18
[Docs][Order] Update docs after removing channel code from order routes

GSadee:
After #465


lchrusciel:
Thank you, Grzegorz! :1st_place_medal:

PR #475, opened 2019-07-22
[POC] Another approach to the command providers

GSadee:
Based on #380


src/CommandProvider/DefaultCommandProvider.php
```diff
@@ -0,0 +1,47 @@
...
+    /** @var ValidatorInterface */
+    private $validator;
+
+    public function __construct(string $requestClass, ValidatorInterface $validator)
+    {
+        $this->requestClass = $requestClass;
+        $this->validator = $validator;
+    }
+
+    protected function transformRequest(Request $request): RequestInterface
```
lchrusciel: Should be below public

src/CommandProvider/DefaultCommandProvider.php
```diff
@@ -0,0 +1,47 @@
...
+namespace Sylius\ShopApiPlugin\CommandProvider;
+
+use Sylius\ShopApiPlugin\Command\CommandInterface;
+use Sylius\ShopApiPlugin\Request\RequestInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Validator\ConstraintViolationListInterface;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
+use Webmozart\Assert\Assert;
+
+class DefaultCommandProvider implements CommandProviderInterface
```
lchrusciel: Do we really need to make it open?

src/Factory/ValidationErrorViewFactory.php
```diff
@@ -29,7 +29,13 @@ public function create(ConstraintViolationListInterface $validationResults): Val
         $errorMessage->message = 'Validation failed';
         /** @var ConstraintViolationInterface $result */
         foreach ($validationResults as $result) {
-            $errorMessage->errors[$result->getPropertyPath()][] = $result->getMessage();
+            // ACL because of new providers path surrounded with `[` and `]`
+            $property = $result->getPropertyPath();
+            if (\preg_match('/\\[.+\\]/', $property) === 1) {
```
lchrusciel: This could be probably removed. My left overs :(

src/CommandProvider/DefaultCommandProvider.php
lchrusciel: And could be private

src/Resources/config/services/command_providers/cart.xml
```diff
@@ -0,0 +1,15 @@
...
+
+<container xmlns="http://symfony.com/schema/dic/services">
+    <services>
+        <defaults public="true" />
+
+        <service
+            id="sylius.shop_api_plugin.command_provider.drop_cart"
+            class="Sylius\ShopApiPlugin\CommandProvider\DefaultCommandProvider"
+        >
+            <argument>Sylius\ShopApiPlugin\Request\Cart\DropCartRequest</argument>
```
JakobTolkemit: I think this could be managed with configuration. Just like the view classes.

src/CommandProvider/DefaultCommandProvider.php
```diff
@@ -0,0 +1,47 @@
...
+        return $this->transformRequest($request)->getCommand();
+    }
+
+    private function transformRequest(Request $request): RequestInterface
+    {
+        $requestModel = call_user_func([$this->requestClass, 'fromRequest'], $request);
+
+        Assert::implementsInterface($requestModel, RequestInterface::class);
+
+        /** @var RequestInterface $requestModel */
```
Zales0123: Shouldn't this be type hinted above line 40?

src/Controller/Cart/PickupCartAction.php
```diff
@@ -36,46 +34,50 @@ final class PickupCartAction
...
     {
+        /** @var ChannelInterface $channel */
         $channel = $this->channelContext->getChannel();
 
-        $pickupRequest = new PickupCartRequest($channel->getCode());
-
-        $validationResults = $this->validator->validate($pickupRequest);
-
+        $validationResults = $this->pickupCartCommandProvider->validate($request, $channel);
         if (0 === count($validationResults)) {
```
Zales0123: The logic inside `if (0 === count($validationResults)) {` is much bigger than after it 💃 we could invert the condition and put more code in the `if` block

src/CommandProvider/DefaultCommandProvider.php
GSadee: IMO this type hint makes sense here, after checking if `$requestModel` implements this interface

lchrusciel:
Thanks, Grzegorz! :tada:

src/CommandProvider/ChannelBasedCommandProvider.php
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    public function getCommand(Request $request, ChannelInterface $channel): CommandInterface
+    {
+        return $this->transformRequest($request, $channel)->getCommand();
+    }
+
+    private function transformRequest(Request $request, ChannelInterface $channel): ChannelBasedRequestInterface
+    {
+        $requestModel = call_user_func([$this->requestClass, 'fromRequestAndChannel'], $request, $channel);
```
CSchulz: why not `$this->requestClass::fromRequest($request)`?

https://www.php.net/manual/en/functions.variable-functions.php

spec/CommandProvider/DefaultCommandProviderSpec.php
```diff
@@ -0,0 +1,38 @@
...
+use Symfony\Component\HttpFoundation\ParameterBag;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Validator\ConstraintViolationListInterface;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
+
+final class DefaultCommandProviderSpec extends ObjectBehavior
+{
+    function let(ValidatorInterface $validator): void
+    {
+        $this->beConstructedWith(DropCartRequest::class, $validator);
```
pamil: I think it'd be better to create an exemplary request for use in the tests, right now this spec will fail if `DropCartRequest` is deleted.

src/Command/CommandInterface.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\Command;
+
+interface CommandInterface
```
pamil: Why wouldn't we stick with `object` typehints instead of `CommandInterface` marker?

src/CommandProvider/ChannelBasedCommandProvider.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        return $this->validator->validate($this->transformRequest($request, $channel));
+    }
+
+    public function getCommand(Request $request, ChannelInterface $channel): CommandInterface
+    {
+        return $this->transformRequest($request, $channel)->getCommand();
+    }
+
+    private function transformRequest(Request $request, ChannelInterface $channel): ChannelBasedRequestInterface
```
pamil: `Request $request` -> `Request $httpRequest`, it's a bit hard to read the code with so many requests that are so different.

src/Request/Cart/DropCartRequest.php
```diff
@@ -5,19 +5,26 @@
...
     protected $token;
 
-    public function __construct(Request $request)
+    private function __construct(Request $request)
     {
         $this->token = $request->attributes->get('token');
     }
 
-    public function getCommand(): DropCart
+    public static function fromRequest(Request $request): RequestInterface
```
pamil: `fromHttpRequest` / `fromSymfonyRequest`?

src/CommandProvider/ChannelBasedCommandProvider.php
CSchulz: Perhaps also change the namespace to HttpRequest

PR #476, opened 2019-07-22
Whitelist ShopApi & LexikJWTAuthenticationBundle integration

lchrusciel:


doc/swagger.yml
```diff
@@ -764,6 +764,27 @@ paths:
...
+                    required: true
+                    schema:
+                        $ref: "#/definitions/LoginRequest"
+            responses:
+                204:
+                    description: "The user was successfully created"
+                400:
+                    description: "There were validation errors"
+                500:
+                    description: "Channel not found"
```
mamazu: Where is the channel specified?

PR #477, opened 2019-07-23
Update README.md

AdamKasp:


README.md
```diff
@@ -18,6 +18,8 @@ The latest documentation is available [here](https://app.swaggerhub.com/apis/Syl
 
 ## Installation
 
+Before install ShopAPI you should disable or delete all ShopBundle dependecy, You can't use these package together !!!
```
Zales0123: Before installing **SyliusShopApiPlugin**, you should disable all **SyliusShopBundle**'s dependencies. You cannot use these packages together.

cc @CoderMaggie

mamazu: Why though?

lchrusciel: ShopApi will not be compatible with ShopBundle at the first stage. We need to customize too many things

mamazu: The only problem I encountered was with the cart provider that the cart indicator of the front-end was not rendering properly what other problems are there?

lchrusciel: Due to the usage of channel context and cart recalculation bug we thought, that it would be better to provide better shop api instead of half baked one

lchrusciel:
Thank you, Adam! :1st_place_medal:

PR #480, opened 2019-07-23
[Product] Refactor product actions to use command providers

GSadee:
Based on #475 and #484


lchrusciel:
Thank you, Grzegorz! :tada:

PR #482, opened 2019-07-23
[Login] Unify and simplify requests

lchrusciel:
Based on #476


GSadee:
Thank you, Łukasz! :1st_place_medal:

PR #484, opened 2019-07-24
Fixes after review for new command providers concept

GSadee:
According to comments in #475

### DIFF
--- a/src/CommandProvider/ChannelBasedCommandProvider.php
+++ b/src/CommandProvider/ChannelBasedCommandProvider.php
@@ -22,6 +22,8 @@ final class ChannelBasedCommandProvider implements ChannelBasedCommandProviderIn
 
     public function __construct(string $requestClass, ValidatorInterface $validator)
     {
+        Assert::implementsInterface($requestClass, ChannelBasedRequestInterface::class);
+
         $this->requestClass = $requestClass;
         $this->validator = $validator;
     }
@@ -38,9 +40,6 @@ final class ChannelBasedCommandProvider implements ChannelBasedCommandProviderIn
 
     private function transformHttpRequest(Request $httpRequest, ChannelInterface $channel): ChannelBasedRequestInterface
     {
-        Assert::methodExists($this->requestClass, 'fromHttpRequestAndChannel');
-        Assert::implementsInterface($this->requestClass, ChannelBasedRequestInterface::class);
-
         /** @var ChannelBasedRequestInterface $request */
         $request = $this->requestClass::fromHttpRequestAndChannel($httpRequest, $channel);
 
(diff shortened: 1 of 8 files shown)


src/CommandProvider/ChannelBasedCommandProvider.php
```diff
@@ -26,23 +26,24 @@ public function __construct(string $requestClass, ValidatorInterface $validator)
...
-        return $this->transformRequest($request, $channel)->getCommand();
+        return $this->transformHttpRequest($httpRequest, $channel)->getCommand();
     }
 
-    private function transformRequest(Request $request, ChannelInterface $channel): ChannelBasedRequestInterface
+    private function transformHttpRequest(Request $httpRequest, ChannelInterface $channel): ChannelBasedRequestInterface
     {
-        $requestModel = call_user_func([$this->requestClass, 'fromRequestAndChannel'], $request, $channel);
+        Assert::methodExists($this->requestClass, 'fromHttpRequestAndChannel');
+        Assert::implementsInterface($this->requestClass, ChannelBasedRequestInterface::class);
```
lchrusciel: Isn't this check enough? Can we skip the previous one?

Also, what would you think about adding this assert to the constructor?

tests/Test/TestChannelBasedCommand.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\Sylius\ShopApiPlugin\Test;
```
lchrusciel:
```suggestion
namespace Tests\Sylius\ShopApiPlugin\Mocks;
```
?

tests/Test/TestChannelBasedCommand.php
```diff
@@ -0,0 +1,32 @@
...
+
+use Sylius\ShopApiPlugin\Command\CommandInterface;
+
+class TestChannelBasedCommand implements CommandInterface
```
lchrusciel:
```suggestion
final class TestChannelBasedCommand implements CommandInterface
```
It is just a test class, we don't need to provide a way to extend it

tests/Test/TestChannelBasedCommand.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /** @var string */
+    protected $token;
```
lchrusciel:
```suggestion
    private $token;
```

Same below

lchrusciel: Or do we even need two arguments here? Let's go with only one

tests/Test/TestChannelBasedRequest.php
```diff
@@ -0,0 +1,35 @@
...
+declare(strict_types=1);
+
+namespace Tests\Sylius\ShopApiPlugin\Test;
+
+use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\ShopApiPlugin\Command\CommandInterface;
+use Sylius\ShopApiPlugin\Request\ChannelBasedRequestInterface;
+use Symfony\Component\HttpFoundation\Request;
+
+class TestChannelBasedRequest implements ChannelBasedRequestInterface
```
lchrusciel:
```suggestion
final class TestChannelBasedRequest implements ChannelBasedRequestInterface
```

lchrusciel:
rebase needed

lchrusciel:
Thank you, Grzegorz! :tada:

PR #485, opened 2019-07-24
[Cart] Remove put multiple items to cart action

GSadee:
I would like to remove this action as this behaviour doesn't exist in Sylius.


mamazu:
Is there a replacement for it? Or would a client have to do multiple `AddToCartRequests` to put multiple items to the cart?

lchrusciel:
The idea was, that there is no such logic in default Sylius distribution. Are you using it in your project?

mamazu:
No but I know of people who use it.

PR #487, opened 2019-07-24
[Customer] Extract customer enabling logic to custom handler

lchrusciel:


src/Handler/Customer/EnableCustomerHandler.php
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    public function __invoke(EnableCustomer $event): void
+    {
+        /** @var ShopUserInterface|null $user */
+        $user = $this->userRepository->findOneByEmail($event->email());
+
+        Assert::notNull($user);
+
+        $user->enable();
```
Zales0123: Don't we need a flush to the database? It was there in the [previous implementation](https://github.com/Sylius/ShopApiPlugin/pull/487/files#diff-6a9800dea25413eea1e73cc7e8207159L56)

lchrusciel: No, as we have doctrine transaction over command bus and it will be flushed there.

Zales0123:
Thank you, Łukasz! :tada:

PR #493, opened 2019-07-25
[Customer] Refactor customer actions to use command providers

GSadee:


src/Controller/Customer/RequestPasswordResettingAction.php
```diff
@@ -24,22 +25,33 @@ final class RequestPasswordResettingAction
...
 
     public function __invoke(Request $request): Response
     {
+        /** @var ChannelInterface $channel */
         $channel = $this->channelContext->getChannel();
 
-        $this->bus->dispatch(new GenerateResetPasswordToken($request->request->get('email')));
-        $this->bus->dispatch(new SendResetPasswordToken($request->request->get('email'), $channel->getCode()));
+        $this->bus->dispatch($this->generateResetPasswordTokenCommandProvider->getCommand($request));
+        $this->bus->dispatch($this->sendResetPasswordTokenCommandProvider->getCommand($request, $channel));
```
lchrusciel: We could also refactor this action to be based on events in next iteration

src/Request/Customer/VerifyAccountRequest.php
```diff
@@ -4,20 +4,27 @@
...
 use Symfony\Component\HttpFoundation\Request;
 
-class VerifyAccountRequest
+class VerifyAccountRequest implements RequestInterface
 {
     /** @var string */
     protected $token;
 
-    public function __construct(Request $request)
+    private function __construct(Request $request)
```
lchrusciel: Quick though. As these classes are meant to be extended, maybe we should define constructors as protected so that they will be usable from child classes.

PR #494, opened 2019-07-25
Display message about decoupling from ShopBundle

Zales0123:
Zrzut ekranu 2019-07-25 o 09 56 32

The message is not-so-pretty, but it could be improved later. cc @lchrusciel @mamazu do you think such a message makes sense, as an addition to [this line](https://github.com/Sylius/ShopApiPlugin/blob/master/README.md#L21) in README 🚀 🐎

### DIFF
--- a/src/DependencyInjection/SyliusShopApiExtension.php
+++ b/src/DependencyInjection/SyliusShopApiExtension.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Sylius\ShopApiPlugin\DependencyInjection;
 
+use Psr\Log\LoggerInterface;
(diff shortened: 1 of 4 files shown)


mamazu:
Isn't there a way to warn the user of it during install. Because otherwise it will just fill the logs every time the container is build.

mamazu:
One thing I would suggest is notifying the user before or after the package is installed (once) and not every time an API call gets made.

You can have a look at the composer package events: https://getcomposer.org/doc/articles/scripts.md#package-events

mamazu:
Could we also maybe completely decouple it from the frontend?

Zales0123:
I believe this PR is not relevant anymore 😄

PR #502, opened 2019-07-30
Add originalPrice to product response

alexander-schranz:
I think the originalPrice is a information which is also mostly needed when reading products.

### DIFF
--- a/spec/Factory/Product/ProductVariantViewFactorySpec.php
+++ b/spec/Factory/Product/ProductVariantViewFactorySpec.php
@@ -80,10 +80,74 @@ final class ProductVariantViewFactorySpec extends ObjectBehavior
         $channel->getBaseCurrency()->willReturn($currency);
         $currency->getCode()->willReturn('PLN');
         $channelPrice->getPrice()->willReturn(500);
+        $channelPrice->getOriginalPrice()->willReturn(null);
 
         $variantView->code = 'SMALL_RED_LOGAN_HAT_CODE';
         $variantView->name = 'Small red Logan hat code';
         $variantView->price = new PriceView();
+        $variantView->axis = ['HAT_SIZE_S', 'HAT_COLOR_RED'];
+        $variantView->nameAxis = [
+            'HAT_SIZE_S' => 'Size S',
+            'HAT_COLOR_RED' => 'Color Red',
+        ];
+
+        $this->create($variant, $channel, 'en_GB')->shouldBeLike($variantView);
+    }
+
+    function it_builds_product_variant_view_with_original_price(
+        PriceViewFactoryInterface $priceViewFactory,
+        ChannelInterface $channel,
+        CurrencyInterface $currency,
+        ChannelPricingInterface $channelPrice,
+        ProductOptionInterface $firstOption,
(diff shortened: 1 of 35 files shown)


spec/Factory/Product/ProductVariantViewFactorySpec.php
```diff
@@ -80,6 +80,7 @@ function it_builds_product_variant_view(
         $channel->getBaseCurrency()->willReturn($currency);
         $currency->getCode()->willReturn('PLN');
         $channelPrice->getPrice()->willReturn(500);
+        $channelPrice->getOriginalPrice()->willReturn(null);
```
mamazu: Should also be tested with originalPrice != null

alexander-schranz: @mamazu test added

src/Factory/Product/ProductVariantViewFactory.php
```diff
@@ -42,6 +42,13 @@ public function create(ProductVariantInterface $variant, ChannelInterface $chann
             $channel->getBaseCurrency()->getCode()
         );
 
+        if ($originalPrice = $channelPricing->getOriginalPrice()) {
```
mamazu: Implicit integer to boolean cast (not type safe)

src/Factory/Product/ProductVariantViewFactory.php
```diff
@@ -42,7 +42,8 @@ public function create(ProductVariantInterface $variant, ChannelInterface $chann
             $channel->getBaseCurrency()->getCode()
         );
 
-        if ($originalPrice = $channelPricing->getOriginalPrice()) {
+        $originalPrice = $channelPricing->getOriginalPrice()
+        if (null !== $originalPrice) {
```
mamazu: The behavior you previously encoded was `$originalPrice !== 0`.

alexander-schranz: If the originalPrice is 0 it should be returned as 0 and only if really not set it should not be returned or?

mamazu: Is it possible for the original price to be null?

alexander-schranz: Yes if not set its null

mamazu: Okay then it okay.

mamazu:
Thank you, Alexander! :tada:

tests/Responses/Expected/order/order_details_response.json
```diff
@@ -33,6 +33,10 @@
                             "current": 1999,
                             "currency": "GBP"
                         },
+                        "originalPrice": {
+                            "current": 2999,
+                            "currency": "GBP"
```
lchrusciel: Is there any possibility to add original price in a different currency than the current one? What would you say, about merging these `originalPrice` and `price` fields?

mamazu: I think both should be used with the same structure. The price is renderd using a `Priceview` so should the original price be otherwise it would make things more difficult when consuming this API.

PR #504, opened 2019-07-30
Moving the complete phpstan analysis to composer

mamazu:
Currently there are two places where phpstan does the checks which can be simpliefied.


lchrusciel:

PR #505, opened 2019-07-31
Returning 401 on customer details/update actions when not logged in

JakobTolkemit:
The `LoggedInCustomerDetailsAction` and the `UpdateCustomerAction` now return an unauthorized response instead of throwing `TokenNotFoundException` if the user is not logged in.


mamazu:
Thanks, Jakob! :tada:

PR #503, opened 2019-07-30
Add onHand to product response

alexander-schranz:
It should be possible to know how many products are still available.

PR #507, opened 2019-07-31
Fix product variant view factory original price test

alexander-schranz:
This fixes the tests on master branch after #502 / #503 where merged into it.


mamazu:

PR #508, opened 2019-08-01
Add availibility property to ProductVariantView

alexander-schranz:
I misinterpredted how sylius checks the availibility of a product when implementing #503. 

This will now use the correct [availibilityChecker](https://github.com/Sylius/Sylius/blob/v1.5.3/src/Sylius/Component/Inventory/Checker/AvailabilityChecker.php) service of sylius.

### DIFF
--- a/spec/Factory/Product/ProductVariantViewFactorySpec.php
+++ b/spec/Factory/Product/ProductVariantViewFactorySpec.php
@@ -10,6 +10,7 @@ use Sylius\Component\Core\Model\ChannelInterface;
 use Sylius\Component\Core\Model\ChannelPricingInterface;
 use Sylius\Component\Core\Model\ProductVariantInterface;
 use Sylius\Component\Currency\Model\CurrencyInterface;
+use Sylius\Component\Inventory\Checker\AvailabilityCheckerInterface;
 use Sylius\Component\Product\Model\ProductOptionInterface;
 use Sylius\Component\Product\Model\ProductOptionTranslationInterface;
 use Sylius\Component\Product\Model\ProductOptionValueInterface;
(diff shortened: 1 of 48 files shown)


alexander-schranz:
Anything missing from my side?

lchrusciel:
No, but both features were added after RC. I'm a little bit worried about adding too much of new functionalities before going to be stable. In default Sylius implementation, we are not presenting the available amount of products atm. 

Also, `onHand` is just a part of the story, because we should take `onHold` into account as well. 

To sum up, I would say, that #503 should be reverted and added after stable release. Probably just by adding `available` flag to products.

AndreasA:
It is still too simple because the availability checker has a second method to check for sufficient quantity. This has to be done when adding a product variant to the cart and when changing its quantity. There exists a constraint `\Sylius\Bundle\CoreBundle\Validator\Constraints\CartItemAvailability` but that one is only useful for `\Sylius\Bundle\OrderBundle\Controller\AddToCartCommandInterface'  but it might be a good reference

AndreasA:
One more suggestion: Instead of a boolean create a state, as users might want to differentiate between `not in stock`, `low on stock` and `available`.
In theory you could even add this information already by providing a low on stock parameter, then:
`isStockAvailable` with `false` would be `not in stock`, `isStockAvailable` with `true` and `isSufficientQuantity` called `parameter value and returns `false` would be `low on stock` and if both are `true` it would be `available`.

mamazu:
I have rebased the merge request to the master. This means that the `onHand` property was removed and replaced with the `available` property. For anyone watching please make sure to check the merge request for any references for the old property. (It was a big rebase with a lot of conflicts so I might have missed one or two instances)

> The base of this pull-request was changed, you need fetch and reset your local branch
execute `git fetch origin && git reset "feature/product-availability-check"` to update your local branch. Feel free to ask for assistance when you get stuck. +1

mamazu:
Thank you, Alexander! :tada:

alexander-schranz:
@mamazu Thank you for rebasing. Didn't got the time to fix the tests 🙈

mamazu:
No problem.

PR #512, opened 2019-08-09
Revert "Add onHand to product response"

lchrusciel:
Reverts Sylius/ShopApiPlugin#503

### DIFF
--- a/doc/swagger.yml
+++ b/doc/swagger.yml
@@ -1321,9 +1321,6 @@ definitions:
                 type: "object"
                 additionalProperties:
                     type: "string"
-            onHand:
-                type: "integer"
-                example: 12
             price:
                 $ref: "#/definitions/Price"
             originalPrice:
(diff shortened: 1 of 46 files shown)


mamazu:
What's the problem with that?

alexander-schranz:
@mamazu As mentioned in #508 we should expose only `availability` flag and not how many are onHand or onHold. As I think onHand and onHold could be also sensitive data for a shop or?

mamazu:
There are two reasons why they should be reverted:
1. the availability is based on more factors than just on hand and on hold quantity.
2. @lchrusciel and I agreed that this might not be part of the 1.1 version as currently we are in a feature freeze phase.

alexander-schranz:
@mamazu
1. yeah thats what I wanted to say.
2. you mean not part of 1.0?

mamazu:
Yes, currently we are just trying to stabalize the current version and not introduce any new features.

mamazu:
Thanks, Łukasz! :tada:

PR #514, opened 2019-08-13
Adding a Cookbook

mamazu:
Well Shop Api got documentation now. Feel free to comment on my not very documentation-like writing style.


doc/Cookbook.md
```diff
@@ -0,0 +1,103 @@
+# Shop Api Plugin
+The Shop Api plugin is a plugin for the Sylius Ecommerce Platform which provides an easy integration for exposing the Sylius functionality to the endcustomer.
```
lchrusciel:
```suggestion
The Shop Api plugin is a plugin for the Sylius Ecommerce Platform which provides an easy integration for exposing the Sylius functionality to the end customer.
```

doc/Cookbook.md
```diff
@@ -0,0 +1,103 @@
...
+        $this->locale = $request->getLocale();
+    }
+
+    public function getCommand(): CommandInterface
+    {
+        return new AddLocalizedCoupon($this->token, $this->coupon, $this->locale);
+    }
+}
+```
+In the same manner the command would need to be created and extending from the `AddCoupon` class. After that the only thing left to do is to is change the request class that is used in the `AddCouponAction` which can be done with configuting the correct request class in the `sylius_shop_api.yml` in the `config` directory.
```
lchrusciel: Now, you should not need to extend an action itself, just override the parameter which contains an FQCN 
https://github.com/Sylius/ShopApiPlugin/blob/master/src/Resources/config/services/command_providers/cart.xml#L11

doc/Cookbook.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Extending handlers
+The main way to extend a handler is to wrap it. This makes it easy to add functionality before and after the handler without the need to change anything. However, if you want to change the logic in the handler you need to overwrite it. This can be done by registering the new Handler with the same service id. **Do not just add it with a new service id** otherwise it will execute both handlers.
+
+### Extending views
+When extending the views there are three places that need to be modified, the view class, the view factory and the view repository (if there are any). The view class is a configuration that can be modified in the `sylius_shop_api.yml` file under the `config` folder. Which in turn gets copied over to a container parameter by the `DependencyInjection/Configuration.php` file where also the request classes are registered.
+
+The ViewFactories as well as the ViewRepositories can be either completely replaced or the prefered way of doing it if you only want to add features to it, decorate the classes.
```
lchrusciel: Could you add a sample decoration?

doc/Cookbook.md
```diff
@@ -0,0 +1,103 @@
+# Shop Api Plugin
+The Shop Api plugin is a plugin for the Sylius Ecommerce Platform which provides an easy integration for exposing the Sylius functionality to the endcustomer.
+
+## Shop Api vs. Admin Api
+In the default implementation of the Sylius Solution there is already an API implemented that provides a lot of features. However this API is more geared towards integrating other closed systems like a warehouse management system. The reason behind that is that for the Api you need to have to exchange tokens to authenticate a new client for usage. In the Shop Api everyone can log into Sylius who has an account, no token exchange necessary.
```
lchrusciel:
```suggestion
In the default implementation of the Sylius Solution, there is already an API implemented that provides a lot of features. However, this API is more geared towards integrating other closed systems like a warehouse management system. The reason behind that is that for the Api you need to have to exchange tokens to authenticate a new client for usage. In the Shop Api everyone can log into Sylius who has an account, no token exchange necessary.
```

doc/Cookbook.md
```diff
@@ -0,0 +1,103 @@
...
+
+## How does the Shop Api work
+<img src="Workflow.png" alt="Apis Workflow" />
+The general approach that Shop Api takes is that every request is validated by the `CommandProvider` and then converted into a command by this class. Then the command is either handled directly in the `Controller` and passed to the `ViewRepository` which returns a view or by a `MessageHandler`.
+
+### The Components
+**Request**: ShopApi has its own request object. This object should abstract away the HTTP Request to a more general request type. Futhermore it also acts as an object that can be validated as all validation rules are defined for the request objects only (commands are not validated)
```
lchrusciel:
```suggestion
**Request**: ShopApi has its own request object. This object should abstract away the HTTP Request to a more general request type. Furthermore, it also acts as an object that can be validated as all validation rules are defined for the request objects only (commands are not validated)
```

doc/Cookbook.md
```diff
@@ -0,0 +1,103 @@
...
+
+**Command**: The command class is an implementation agnostic class that holds the relevant data for handling the command.
+
+**Hanlder**: The handler is the class that defines the logic what happens when a certain command is called. Here we have the buisness logic.
```
lchrusciel:
```suggestion
**Handler**: The handler is the class that defines the logic of what happens when a certain command is called. Here we have the business logic.
```

doc/Cookbook.md
```diff
@@ -0,0 +1,103 @@
...
+
+**ViewFactory**: The view factories are responsible for converting the entities into views.
+
+**Views**: Views are primitive object that only hold scalars or other View objects which can be easily serialzied.
```
lchrusciel:
```suggestion
**Views**: Views are a primitive object that only holds scalars or other View objects which can be easily serialized.
```

doc/Cookbook.md
```diff
@@ -0,0 +1,103 @@
...
+
+**ViewRepository**: ViewRepositories are repositories that return view objects from the entities they are being fetchted
```
lchrusciel:
```suggestion
**ViewRepository**: ViewRepositories are repositories that return view objects from the entities they are fetched.
```

doc/Cookbook.md
```diff
@@ -0,0 +1,103 @@
...
+
+> ViewRepositories are **not** repositories of views.
+
+### Command - Handler Structure
+The Command handling has multiple parts to it. When dispatching a command the `MessageBus` looks for a handler that has an `__invoke` method with the parameter type that matches the type of the command that was dispatched. Before and after the handler is executed there is a way for a "Middleware" to be executed (see below). The CommandHandler itself however doesn't return anything (this is not a technical limitation that is just the convention we chose in ShopApi).
```
lchrusciel:
```suggestion
Command handling has multiple parts to it. When dispatching a command, the `MessageBus` looks for a handler that has an `__invoke` method with the parameter type that matches the type of the command that was dispatched. Before and after the handler is executed, there is a way for a "Middleware" to be executed (see below). The CommandHandler itself, however, doesn't return anything (this is not a technical limitation that is just the convention we chose in ShopApi).
```

doc/Cookbook.md
```diff
@@ -0,0 +1,103 @@
...
+
+> Important: The enties that were loaded from the `EntityManager` and changed in the `CommandHandler` are flushed in the Middleware. More information about the Middleware [here](https://symfony.com/doc/current/components/messenger.html#bus).
```
lchrusciel:
```suggestion
> Important: The entities that were loaded from the `EntityManager` and changed in the `CommandHandler` are flushed in the Middleware. More information about the Middleware [here](https://symfony.com/doc/current/components/messenger.html#bus).
```

doc/Cookbook.md
```diff
@@ -0,0 +1,103 @@
...
+    public function getCommand(): CommandInterface
+    {
+        return new AddLocalizedCoupon($this->token, $this->coupon, $this->locale);
+    }
+}
+```
+In the same manner the command would need to be created and extending from the `AddCoupon` class. After that the only thing left to do is to is change the request class that is used in the `AddCouponAction` which can be done with configuting the correct request class in the `sylius_shop_api.yml` in the `config` directory.
+
+### Extending handlers
+The main way to extend a handler is to wrap it. This makes it easy to add functionality before and after the handler without the need to change anything. However, if you want to change the logic in the handler you need to overwrite it. This can be done by registering the new Handler with the same service id. **Do not just add it with a new service id** otherwise it will execute both handlers.
```
lchrusciel:
```suggestion
The main way to extend a handler is to wrap it. This makes it easy to add functionality before and after the handler without the need to change anything. However, if you want to change the logic in the handler, you need to overwrite it. This can be done by registering the new handler with the same service id. **Do not just add it with a new service id** otherwise, it will execute both handlers.
```

doc/Cookbook.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Extending views
+When extending the views there are three places that need to be modified, the view class, the view factory and the view repository (if there are any). The view class is a configuration that can be modified in the `sylius_shop_api.yml` file under the `config` folder. Which in turn gets copied over to a container parameter by the `DependencyInjection/Configuration.php` file where also the request classes are registered.
```
lchrusciel:
```suggestion
When extending the views, three places need to be modified, the view class, the view factory, and the view repository (if there are any). The view class is a configuration that can be modified in the `sylius_shop_api.yml` file under the `config` folder. Which in turn gets copied over to a container parameter by the `DependencyInjection/Configuration.php` file where also the request classes are registered.
```

doc/Cookbook.md
lchrusciel:
```suggestion
The ViewFactories, as well as the ViewRepositories, can be either completely replaced or the preferred way of doing it if you only want to add features to it, decorate the classes.
```

doc/Cookbook.md
```diff
@@ -0,0 +1,103 @@
...
+
+The ViewFactories as well as the ViewRepositories can be either completely replaced or the prefered way of doing it if you only want to add features to it, decorate the classes.
+
+### Custom Channel Handlers
+The default way that Sylius tries to resolve channels is through the hostname. However if you want to add your own way of resolving a channel for example from a route parameter you need to do two things:
```
lchrusciel:
```suggestion
The default way that Sylius tries to resolve channels is through the hostname. However, if you want to add your own way of resolving a channel, for example from a route parameter, you need to do two things:
```

mamazu:
Adding information about Serialization

doc/Cookbook.md
mamazu: Yeah, that's what I meant. I probably need to reword that. :D

doc/Cookbook.md
```diff
@@ -0,0 +1,147 @@
+# Shop Api Plugin
+The Shop Api plugin is a plugin for the Sylius Ecommerce Platform which provides an easy integration for exposing the Sylius functionality to the end customer.
```
CSchulz: Perhaps add here also a mobile app or similar.

doc/Cookbook.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Shop Api vs. Admin Api
+In the default implementation of the Sylius Solution, there is already an API implemented that provides a lot of features. However, this API is more geared towards integrating other closed systems like a warehouse management system. The reason behind that is that for the Api you need to have to exchange tokens to authenticate a new client for usage. In the Shop Api everyone can log into Sylius who has an account, no token exchange necessary.
```
CSchulz: To be more clear add Admin API here.

CSchulz: Link to the token documentation?

doc/Cookbook.md
```diff
@@ -0,0 +1,147 @@
...
+
+## How does the Shop Api work
+<img src="Workflow.png" alt="Apis Workflow" />
+The general approach that Shop Api takes is that every request is validated by the `CommandProvider` and then converted into a command by this class. Then the command is either handled directly in the `Controller` and passed to the `ViewRepository` which returns a view or by a `MessageHandler`.
```
CSchulz: I would suggest to add here some links to the different variants. I think the `MessageHandler` is part of the Messenger framework, right?

doc/Cookbook.md
```diff
@@ -0,0 +1,147 @@
...
+**Views**: Views are primitive objects that only hold scalars or other View objects which can be easily serialzied.
+
+**ViewRepository**: ViewRepositories are repositories that return view objects from the entities they are fetched.
+
+> ViewRepositories are **not** repositories of views.
+
+### Command - Handler Structure
+Command handling has multiple parts to it. When dispatching a command, the `MessageBus` looks for a handler that has an `__invoke` method with the parameter type that matches the type of the command that was dispatched. Before and after the handler is executed, there is a way for a "Middleware" to be executed (see below). The CommandHandler itself, however, doesn't return anything (this is not a technical limitation that is just the convention we chose in ShopApi).
+
+> Important: The entities that were loaded from the `EntityManager` and changed in the `CommandHandler` are flushed in the Middleware. More information about the Middleware [here](https://symfony.com/doc/current/components/messenger.html#bus).
```
CSchulz: What do you mean with flush? The link doesn't contain any hint for flush.

doc/Cookbook.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Extending Shop Api
+
+### Extending requests / commands
+The easiest way to extend a Request object and add new properties is to extend from the Shop Api Object. There you need to change the constructor to extract the additional infomration from the request and then also override the corresponding command class. For example adding a locale to the coupon:
```
CSchulz: > to extend the existing request classes.

doc/Cookbook.md
```diff
@@ -0,0 +1,147 @@
...
+```
+In the same manner the command would need to be created and extending from the `AddCoupon` class. To change the flow from using the Shop Api request and command object to your own structure you only need to change one parameter in the configuration under the `sylius_shop_api.yml` like so:
+```yml
+sylius_shop_api:
+    request_classes:
+        add_coupon: \AddLocalizedCouponRequest
+```
+
+### Extending handlers
+The main way to extend a handler is to wrap it. This makes it easy to add functionality before and after the handler without the need to change anything. However, if you want to change the logic in the handler, you need to overwrite it. This can be done by registering the new handler with the same service id. **Do not just add it with a new service id** otherwise, it will execute both handlers.
```
CSchulz: Use for the bold text also an cite like above for the important text.

CSchulz: Can we also decorate an existing handler?

doc/Cookbook.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Extending views
+When extending the views, three places need to be modified, the view class, the view factory, and the view repository (if there are any). The view class is a configuration that can be modified in the `sylius_shop_api.yml` file under the `config` folder. Which in turn gets copied over to a container parameter by the `DependencyInjection/Configuration.php` file where also the request classes are registered.
```
CSchulz: Do we need the part with the `Configuration.php`?
I thought this is the general part down by the DI and nothing Sylius specific, right?

doc/Cookbook.md
```diff
@@ -0,0 +1,147 @@
...
+
+        $totalView->nicePersonDiscount = $orderInterface->getNiceDiscount();
+
+        return $totalView;
+    }
+}
+```
+and this is configured like this:
+```xml
+<service class="NiceTotalView"
```
CSchulz: yaml isn't possible?

doc/Cookbook.md
```diff
@@ -0,0 +1,147 @@
...
+
+**Handler**: The handler is the class that defines the logic of what happens when a certain command is called. Here we have the business logic.
+
+**ViewFactory**: The view factories are responsible for converting the entities into views.
+
+**Views**: Views are primitive objects that only hold scalars or other View objects which can be easily serialzied.
+
+**ViewRepository**: ViewRepositories are repositories that return view objects from the entities they are fetched.
+
+> ViewRepositories are **not** repositories of views.
```
CSchulz: I am not sure what you want to say.

doc/Cookbook.md
mamazu: Yes in the MessageHandler section there is a link to the symfony configuration of this tool.

mamazu:
@CSchulz Thanks for your feedback, having no knowledge about the api is the best thing, if you are reading through documentation. I have improved on your suggestions. Could you have a look at it again.

CSchulz:
Yes I will look into it at the weekend.

doc/Cookbook.md
```diff
@@ -0,0 +1,153 @@
...
+        total: \NiceTotalView
+```
+
+### Custom Channel Handlers
+The default way that Sylius tries to resolve channels is through the hostname. However, if you want to add your own way of resolving a channel, for example from a route parameter, you need to do two things:
+
+1. Create a class that resolves the URL to a channel
+
+```php
+class RequestAttributeChannelContext implements ChannelContext
```
diimpp: `ChannelContextInterface`

Shouldn't be there `use`'s for classes in use in this example?

Personally, I think namespace should be specified as well, otherwise people will start adding them as is, seen too many times. For example `namespace App\ShopApi\Context`.

mamazu: True. I will also add that to the documentation.

lchrusciel:
If rebased could be probably merged.

doc/Cookbook.md
```diff
@@ -0,0 +1,157 @@
+# Shop Api Plugin
+The Shop Api Plugin is a plugin for the Sylius E-Commerce Platform which provides an easy integration for exposing the Sylius functionality to the end customer. One use-case would be if you want to run your shop without a frontend and maybe want to display the products and handle the cart flow in a mobile app. With this Plugin you just need to send simple Rest Requests to interact with Sylius.
+
+## Shop Api vs. Admin Api
+In the default implementation of the Sylius Solution, there is already an api implemented that provides a lot of features (which is called the Admin Api). The Admin Api is more geared towards integrating other closed systems like a warehouse management system or similar. The reason behind that is that for the Admin Api you need to have to [exchange tokens to authenticate](https://docs.sylius.com/en/latest/cookbook/api/api.html) a new client for usage (which is based on oauth). On the other side in the Shop Api everyone can log into Sylius who has an account, no token exchange necessary. The Shop Api uses the [JWT](https://github.com/lexik/LexikJWTAuthenticationBundle) to authenticate its users.
```
CSchulz: Just a wording thing:

The default implementation provided by Sylius Solution already implements an api containing a lot of features (called Admin Api)

doc/Cookbook.md
```diff
@@ -0,0 +1,157 @@
...
+
+### The Components
+* **Request**: The Shop Api has its own request object. This object should abstract away the HTTP Request to a more general request type. Furthermore, it also acts as an object that can be validated as all validation rules are defined for the request objects only (commands are not validated)
+* **Command**: The command class is an implementation agnostic class that holds the relevant data for handling the command.
+* **Handler**: The handler is the class that defines the logic of what happens when a certain command is called. Here we have the business logic.
+* **ViewFactory**: The view factories are responsible for converting the entities into views.
+* **Views**: Views are primitive objects that only hold scalars or other View objects which can be easily serialized.
```
CSchulz: Perhaps you can add there POPO (plain old php object).

doc/Cookbook.md
```diff
@@ -0,0 +1,157 @@
...
+
+## How does the Shop Api work
+<img src="Workflow.png" alt="Apis Workflow" />
+The general approach that Shop Api takes is that every request, validates it and turns it into a command with the `CommandProvider`. Then this command is either handled directly in the `Controller` (in case of a get route for example) and passed to the `ViewRepository` which returns a view. For requests that should change the system's state the command is dispatched and taken care of by one of the `MessageHandler`s.
```
CSchulz: I am not sure where it belongs, but I miss the part where the static method of the `RequestInterface` is called.

doc/Cookbook.md
```diff
@@ -0,0 +1,157 @@
...
+
+### The Components
+* **Request**: The Shop Api has its own request object. This object should abstract away the HTTP Request to a more general request type. Furthermore, it also acts as an object that can be validated as all validation rules are defined for the request objects only (commands are not validated)
```
CSchulz: In general I am not sure if we should better rename the Request class / interface to remove confusion.

doc/Cookbook.md
```diff
@@ -0,0 +1,157 @@
...
+
+## How does the Shop Api work
+<img src="Workflow.png" alt="Apis Workflow" />
+The general approach that Shop Api takes is that every request, validates it and turns it into a command with the `CommandProvider`. Then this command is either handled directly in the `Controller` (in case of a get route for example) and passed to the `ViewRepository` which returns a view. For requests that should change the system's state the command is dispatched and taken care of by one of the `MessageHandler`s.
+
+### The Components
```
CSchulz: All the component descriptions could be part of the PHPDoc of the right interfaces.

doc/Cookbook.md
CSchulz: > takes is that every request

I think that are the `Action` classes, aren't they? Perhaps add here the hint.

mamazu: Which static method?

doc/Cookbook.md
mamazu: Do you have any suggestions?

doc/Cookbook.md
mamazu: I didn't know that Action classes was a thing. I can add the word if that helps.

doc/Cookbook.md
mamazu: Well the handlers don't have any common interface, the request and command objects either so there is no real other place for it.

doc/Cookbook.md
```diff
@@ -0,0 +1,157 @@
...
+
+    public function __construct(ChannelRepositoryInterface $channelRepository, RequestStack $requestStack)
+    {
+        $this->channelRepository = $channelRepository;
+        $this->requestStack = $requestStack;
+    }
+
+    public function getChannel(): ChannelInterface
+    {
+        $request = $this->requestStack->getCurrentRequest();
```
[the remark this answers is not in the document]
mamazu: The reason why this case is documented is that in the past the routes all had a channel attribute so this is more a documentation on how to maintain backwards compatibiltiy rather than best practice. Best practice is to use the hostname to resolve the channel.

doc/Cookbook.md
```diff
@@ -0,0 +1,157 @@
...
+```xml
+<service id="sylius.context.channel.request_attribute_based" class="RequestAttributeChannelContext">
+    <argument type="service" id="sylius.repository.channel" />
+    <argument type="service" id="request_stack" />
+
+    <tag name="sylius.context.channel" priority="100" />
+</service>
+```
+### Handling Payments via API
+**Currently the API does not support any kind of payments.** A simple way to implement this is to define an endpoint that will take the payment data and process it in Sylius. One approach here is to make all payments in Sylius "offline payments" so that the api can handle this manually.
```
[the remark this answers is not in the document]
mamazu: There is still no agreement on how to correctly implement a payment with the API because everybody has different requirements.

doc/Cookbook.md
```diff
@@ -0,0 +1,157 @@
...
+and to change the view class in sylius:
+```yml
+sylius_shop_api:
+    view_classes:
+        total: \NiceTotalView
+```
+
+### Custom Channel Handlers
+The default way that Sylius tries to resolve channels is through the hostname. However, if you want to add your own way of resolving a channel, for example from a route parameter, you need to do two things:
+
```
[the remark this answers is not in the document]
mamazu: True there might also be a link to the documentation of the pagerfanta serialization.

diimpp:
I wouldn't say, that hostname resolving is a best practice, just sensible
default for the plugin.

My point is that this example is full of errors, it would be better to
follow one of the Core examples.

mamazu:
True, Sylius Core handles it that way so this plugin should allow suit for the hostname resolution.
I have updated the documentation, any thoughts?

doc/Cookbook.md
CSchulz: Can you explain it, please?

doc/Cookbook.md
CSchulz: `RequestInterface::fromHttpRequest` which is used to create the Request instance.

CSchulz: I meant f.e. https://github.com/Sylius/ShopApiPlugin/blob/9007898c08da965c1e163209244f219af2ee6e95/src/Controller/Product/AddReviewByCodeAction.php

doc/Cookbook.md
CSchulz: Request: RequestInterface
Command: CommandInterface
Handler: no interface available (yet)
View: no interface available (yet)

doc/Cookbook.md
CSchulz: I was thinking about ApiRequest or ApiPayload or RequestPayload.

doc/Cookbook.md
```diff
@@ -0,0 +1,170 @@
...
+## Shop Api vs. Admin Api
+In the default implementation of the Sylius Solution, there is already an api implemented that provides a lot of features (which is called the Admin Api). The Admin Api is more geared towards integrating other closed systems like a warehouse management system or similar. The reason behind that is that for the Admin Api you need to have to [exchange tokens to authenticate](https://docs.sylius.com/en/latest/cookbook/api/api.html) a new client for usage (which is based on oauth). On the other side in the Shop Api everyone can log into Sylius who has an account, no token exchange necessary. The Shop Api uses the [JWT](https://github.com/lexik/LexikJWTAuthenticationBundle) to authenticate its users.
+
+## How does the Shop Api work
+<img src="Workflow.png" alt="Apis Workflow" />
```
CSchulz: I am wondering why the request isn't mentioned in the workflow, perhaps you can also add it to the beginning.

doc/Cookbook.md
mamazu: True that's why I didn't want to have the documentation spread across the whole project instead here in one place.

doc/Cookbook.md
mamazu: This is called in the command provider.

mamazu:
This is the last iteration of the cookbook before 1.0

doc/Cookbook.md
CSchulz: Yes, I am not sure, but I think there is a small part missing explaining this relationship.

doc/Cookbook.md
```diff
@@ -0,0 +1,170 @@
...
+* **Views**: Views are primitive objects that only hold scalars or other View objects which can be easily serialized (plain old php objects).
+* **ViewRepository**: ViewRepositories are repositories that return view objects from the entities they are fetched.
+
+> ViewRepositories are **not** repositories of views which means they don't save views and can not be used for caching.
+
+### Command - Handler Structure
+Command handling has multiple parts to it. When dispatching a command, the `MessageBus` looks for a handler that has an `__invoke` method with the parameter type matching the type of the command that was dispatched. Before and after the handler is executed, there is a way for a "Middleware" to be executed (see below). The CommandHandler itself, however, does not return anything (this is not a technical limitation that is just the convention we chose in Shop Api).
+
+All Middlewares which are executed are defined under the `framework.messenger` bundle: [config.yml](https://github.com/Sylius/ShopApiPlugin/blob/fc25f36274e6add118f5b575a44db81bcc47b2e5/src/Resources/config/app/config.yml#L17)
```
lchrusciel:
```suggestion
All Middlewares which are executed are defined under the `framework.messenger` bundle: [config.yml](https://github.com/Sylius/ShopApiPlugin/blob/master/src/Resources/config/app/config.yml#L17)
```

lchrusciel:

PR #520, opened 2019-08-24
docs(swagger): fix `/carts/{token}/multiple-items`

Gounlaf:
Class `src/Controller/Cart/PutItemsToCartAction.php#L75` expect items to in an `items` property from the payload


mamazu:
Thank you, Florian! :1st_place_medal:

Issue #9, opened 2017-03-24
As a Customer, I want to reset my password


pjedrzejewski:
* Set new password using a token

```
PUT /password-reset/{token}
```

lchrusciel:
Fixed by #143

PR #525, opened 2019-09-02
Add route for changing password of current user

### LINKED ISSUE #28: As a Customer, I want to change my password
* Change password (only authenticated)

```
PUT /change-password
```

Parameters:

* Current password
* New password
* Verification

nnatter:
Closes #28

This PR adds a `sylius_shop_api_change_password` route with the path `/me/change-password`.
The route allows to change the password of the user that is currently logged in.

The PR does not implement any new action, but rather uses the existing `changePasswordAction` of the `UserController` of Sylius. This is the same strategy that is already used for the `sylius_shop_api_password_reset` route.

### DIFF
--- a/tests/Controller/Customer/ChangePasswordApiTest.php
+++ b/tests/Controller/Customer/ChangePasswordApiTest.php
@@ -35,6 +35,19 @@ JSON;
 
         $response = $this->client->getResponse();
         $this->assertResponseCode($response, Response::HTTP_NO_CONTENT);
+
+        // test login with new password
+        $newLoginData =
+<<<JSON
+        {
+            "email": "oliver@queen.com",
+            "password": "new-pass"
+        }
+JSON;
+        $this->client->request('POST', '/shop-api/login', [], [], ['CONTENT_TYPE' => 'application/json', 'ACCEPT' => 'application/json'], $newLoginData);
+
+        $response = $this->client->getResponse();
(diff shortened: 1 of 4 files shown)


tests/Controller/Customer/ChangePasswordApiTest.php
```diff
@@ -0,0 +1,88 @@
...
+                "first": "new-pass",
+                "second": "new-pass"
+            }
+        }
+JSON;
+
+        $this->client->request('PUT', '/shop-api/me/change-password', [], [], self::CONTENT_TYPE_HEADER, $changePasswordData);
+
+        $response = $this->client->getResponse();
+        $this->assertResponseCode($response, Response::HTTP_NO_CONTENT);
```
mamazu: Maybe we should try to log in after changing the password to assert that it really works.

lchrusciel:
Hey @nnatter,

thanks a lot for the contribution! 

@mamazu as we don't add any new code here, maybe we should add it before 1.0? On the one hand, the risk is rather low, as the feature existed already and it is tested. On the other hand, we have feature freeze before the stable release.

mamazu:
Yeah, quite a basic feature but as you said feature freeze. I would wait for 1.1.

mamazu:
Closes #9

mamazu:
Closes #28

mamazu:
Thanks, @nnatter! :1st_place_medal:

PR #527, opened 2019-09-03
Remove trailing slashes from route paths

nnatter:
This PR removes all remaining trailing slashes from the paths of the API routes. 

This is mainly a consistency thing as these routes are documented without trailing slashes in the `doc/swagger.yml` file.
Furthermore, symfony should redirect these paths anyway 🙂 (see https://symfony.com/doc/4.1/routing.html#redirecting-urls-with-trailing-slashes).

### DIFF
--- a/src/Resources/config/routing/address_book.yml
+++ b/src/Resources/config/routing/address_book.yml
@@ -1,5 +1,5 @@
 sylius_shop_api_address_book_create:
-    path: /address-book/
+    path: /address-book
     controller: sylius.controller.address:createAction
     methods: [POST]
     defaults:
@@ -10,7 +10,7 @@ sylius_shop_api_address_book_create:
                     - "expr:service('sylius.context.customer').getCustomer()"
 
 sylius_shop_api_address_book_index:
-    path: /address-book/
+    path: /address-book
     controller: sylius.controller.address:indexAction
     methods: [GET]
     defaults:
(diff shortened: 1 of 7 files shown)


mamazu:
@lchrusciel This should be part of version 1.0 as well I think.

lchrusciel:
Make sense, especially if it is Symfony recommendation.

However, this library is supporting Symfony 3.4, which will make an inconsistency in paths resolution if someone has implemented the communication with trailing slashes.

We need to add this to UPGRADE file then, as it will be a BC-Break

mamazu:
Yes, if it is documented I would also agree that it is a good thing.

lchrusciel:
After a deeper thougt, I think we should adjust documentation to existing code.

kortwotze:
So this will not be merged?

mamazu:
Yes because although Symfony handles the trailing slashes it creates a redirect to the correct route which is not in line with what the documentation states. So the documentation should document the routes that have to be triggered correctly.

mamazu:
I think this is something for Version 1.1

lchrusciel:
I thought, we've agreed that rather than changing routing we should adjust documentation

mamazu:
Agreed.

nnatter:
Closing this to clean up my GitHub dashboard.

PR #528, opened 2019-09-04
Fix CartBlamerListener case with empty token

diimpp:
This PR solves false-positive CartBlamerListener execution.

> $cart = $this->cartRepository->findOneBy(['tokenValue' => $token]);

If database contains orders without `tokenValue`, then it will return non-null `$cart` result for null `$token`. [Order.orm.xml tokenValue is nullable](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/Order.orm.xml#L76)

### DIFF
--- a/src/EventListener/CartBlamerListener.php
+++ b/src/EventListener/CartBlamerListener.php
@@ -44,7 +44,9 @@ final class CartBlamerListener
             return;
         }
 
-        $token = $request->request->get('token');
+        if (null === $token = $request->request->get('token')) {
+            return;
+        }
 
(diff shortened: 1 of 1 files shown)


mamazu:
One thing I would even recommend is using a CartProvider with token. This would also allow users to add more constraints and checks on it (eg. is cart in current channel and so on).

lchrusciel:
Thanks Dmitri for the contribution! Nevertheless, this issue will be solved by #536, as it contains a documentation as well

PR #530, opened 2019-09-05
Addding assertions for product in current channel

### LINKED ISSUE #59: Adding a product from different channel.
Right now it is possible to add a product to the cart even if this product is not available for given channel

mamazu:
Fixes #59 

## What does it do?
Currently you can add any product to any cart. After this pull request the product needs to be available in the cart's channel. So that if a product is in channel A and the cart is picked up in channel B then the product-add endpoint will now throw an error (500).

## Why a 500 and not a 400 with validation errors?
The problem is that the Validation would need to take two properties of the request into consideration: the current cart token, the product code. Which means that it has to be a "Class Constraint" and needs to access properties of the request object. This is currently not possible as the current request objects don't have getter and the properties are private. So there is no way of validating them. Therefore this pull request only adds an assertion that it is true. (validation may follow if it will be possible to do so in the future)

PS: @JakobTolkemit you might be interested.

### DIFF
--- a/doc/swagger.yml
+++ b/doc/swagger.yml
@@ -141,6 +141,8 @@ paths:
                     description: "Invalid input, validation failed."
                     schema:
                         $ref: "#/definitions/GeneralError"
+                500:
+                    description: "An error occurred - maybe the product is not in the channel of the cart"
 
     /carts/{token}/multiple-items:
         parameters:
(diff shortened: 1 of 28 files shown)


lchrusciel:
We should fix it before 1.0. Nevertheless, I would vote for opening Request object, as this solution would be much cleaner.

mamazu:
How would that work though?

lchrusciel:
Thank you, @mamazu! :tada:

PR #531, opened 2019-09-06
Fix cors configuration and paths to config file

alexander-schranz:
`sylius_shop_api_set_default_address` is a PATCH request


lchrusciel:
Thank you, Alexander! :tada:

PR #536, opened 2019-09-16
Fix CartBlamerListener without token

alexander-schranz:
Carts which are not created over the ShopApi have no tokenValue so if you try to login it will accidently try to assign a Cart which is not the one of the current value.

So a cart should only tried to be loaded when the $token is given to the api.


mamazu:

PR #542, opened 2019-09-17
fix: return array for product select attribute type, improve testing …

denis019:
There was a bug when you have product with multi select attribute, only the first one was returned.
I have adapt the tests for the multi language select attributes since was missing.

### DIFF
--- a/src/Factory/Product/ProductAttributeValueViewFactory.php
+++ b/src/Factory/Product/ProductAttributeValueViewFactory.php
@@ -27,7 +27,7 @@ final class ProductAttributeValueViewFactory implements ProductAttributeValueVie
         $productAttributeValueView->code = $productAttributeValue->getCode();
 
         if ($productAttributeValue->getType() === 'select') {
-            $productAttributeValueView->value = $this->resolveSelectAttribute($productAttributeValue);
+            $productAttributeValueView->value = $this->resolveSelectAttribute($productAttributeValue, $locale);
         } else {
             $productAttributeValueView->value = $productAttributeValue->getValue();
         }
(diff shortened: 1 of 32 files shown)


mamazu:
Thanks for your contribution. I would suggest splitting up the merge request into two parts. One with the tests and one with the new feature.

denis019:
> Thanks for your contribution. I would suggest splitting up the merge request into two parts. One with the tests and one with the new feature.

@mamazu thanks for checking this PR, to me this is a fix for the select attribute, because the multi select use case was missing.
Tests are related with the PR because the select will always return array.
Let me know your opinion.

mamazu:
Ah now I see where you are going. Then this makes a lot more sense. The only problem I still have with it is the fact that the response of the ShopApi has to be localized in the language of the locale.

tests/Responses/Expected/cart/add_multiple_products_to_cart_response.json
```diff
@@ -45,7 +45,12 @@
                     {
                         "code": "MUG_MATERIAL_CODE",
                         "name": "Mug material",
-                        "value": "Wood"
+                        "value": [
```
lchrusciel: Value returned to UI should be localized. Therefore, the previous returned value was ok

src/Factory/Product/ProductAttributeValueViewFactory.php
```diff
@@ -43,14 +43,16 @@ public function create(ProductAttributeValueInterface $productAttributeValue, st
...
-        if ($choices === null || $value === null) {
-            return null;
+        foreach ($productAttributeValue->getValue() as $value) {
+            if (array_key_exists($value, $choices)) {
+                $values[] = $choices[$value];
+            }
         }
 
-        return $choices[$value];
+        return $values;
```
lchrusciel: It makes sense to return an array but only, when the attribute value contains more than one field. Translations itself should be handled by ShopAPI,

tests/Responses/Expected/cart/german_filled_cart_with_simple_product_summary_response.json
```diff
@@ -40,7 +40,9 @@
                     {
                         "code": "MUG_MATERIAL_CODE",
                         "name": "Becher Material",
-                        "value": "Wood"
+                        "value": [
```
lchrusciel: I was thinking about how to solve it. What do you think about something like that:
```suggestion
                        "type": list
                        "value": [
```
and 
```suggestion
                        "type": scalar
                        "value": "Wood"
```

And later it could be extended by custom types, like object or whatever

lchrusciel:
Closing in favour of #586

PR #546, opened 2019-09-18
Handle registration for exist customer

### LINKED ISSUE #533: 500 error when registering with exist email address
When registering with exist email address it should return a 400 status code and not a server side error.

alexander-schranz:
Closes #533

This would handle the registration process like Standard Sylius that a Customer is set to the User if the E-Mail Address exist as a customer. Currently you are not able to register if you did an anonymous order over the shop api.

### DIFF
--- a/src/Handler/Customer/RegisterCustomerHandler.php
+++ b/src/Handler/Customer/RegisterCustomerHandler.php
@@ -7,6 +7,7 @@ namespace Sylius\ShopApiPlugin\Handler\Customer;
 use Sylius\Component\Channel\Repository\ChannelRepositoryInterface;
 use Sylius\Component\Core\Model\CustomerInterface;
 use Sylius\Component\Core\Model\ShopUserInterface;
+use Sylius\Component\Core\Repository\CustomerRepositoryInterface;
 use Sylius\Component\Resource\Factory\FactoryInterface;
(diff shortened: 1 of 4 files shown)


mamazu:
What happens if I enter an already existing email can I change the whole customer with that. This would mean I can just "register" with my friends email address and take over his account.

alexander-schranz:
@mamazu like in Sylius you first need to click on the Confirmation Link in the email to enable the new created account so without access to the Email account you will not be able to login or see anything in this account.

robmeek:
@alexander-schranz  
I’m doing this same thing in my app.
But might it be better not to change any fields on the customer entity if the customer exists – so you put $customer->setFirstName etc inside the conditional.
 – otherwise, while you might not be able to hijack an account because of the confirmation email, you could easily change one’s name.

alexander-schranz:
@robmeek Thats a good point! Sadly this is what sylius currently also does when you do it over the normal sylius standard UI I think this should then also be changed there.

//cc @lchrusciel

lchrusciel:
Totally makes sense to add this behavior. As close as possible to Sylius I would say. It could be considered as a bug even.

lchrusciel:
Fixes #533

mamazu:

PR #548, opened 2019-09-21
Fixing the customer birthday

### LINKED ISSUE #544: Updating a customer without a birthday sets the birthday to the current dat
When the `birthday` property is omitted (or not set) on a PUT request to the `/me` endpoint, the birthday of the current customer is set to the current date.

The `UpdateCustomerRequest` class declares the `birthday` class-property as nullable, therefore I assume that it should be possible to omit the `birthday` property from this request (eg. if the customer does not want to specify his birthday).
https://github.com/Sylius/ShopApiPlugin/blob/master/src/Request/Customer/UpdateCustomerRequest.php#L25

This behaviour happens because the `UpdateCustomerHandler` constructs a new `DateTimeImmutable` object with the `birthday` property of the `UpdateCustomerRequest`. If this property is null, the constructed `DateTimeImmutable` object will contain the current date.
https://github.com/Sylius/ShopApiPlugin/blob/master/src/Handler/Customer/UpdateCustomerHandler.php#L31

mamazu:
Fixes #544 
Fixing the customer birthday

### DIFF
--- a/src/Command/Customer/UpdateCustomer.php
+++ b/src/Command/Customer/UpdateCustomer.php
@@ -55,9 +55,13 @@ class UpdateCustomer implements CommandInterface
         return $this->email;
     }
 
-    public function birthday(): ?string
+    public function birthday(): ?\DateTimeImmutable
     {
-        return $this->birthday;
+        if ($this->birthday === null) {
+            return null;
+        }
+
+        return new \DateTimeImmutable($this->birthday);
     }
 
     public function gender(): ?string
--- a/src/Command/Customer/UpdateCustomer.php
+++ b/src/Command/Customer/UpdateCustomer.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Sylius\ShopApiPlugin\Command\Customer;
 
+use DateTimeImmutable;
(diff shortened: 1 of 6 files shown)


src/Handler/Customer/UpdateCustomerHandler.php
```diff
@@ -28,7 +28,7 @@ public function __invoke(UpdateCustomer $command): void
         $customer->setLastName($command->lastName());
         $customer->setEmail($command->email());
         $customer->setGender($command->gender());
-        $customer->setBirthday(new \DateTimeImmutable($command->birthday()));
+        $customer->setBirthday($command->birthday());
```
lchrusciel: What would you say about returning whatever is passed to birthday and moving an `if` statement to handler class?

src/Command/Customer/UpdateCustomer.php
```diff
@@ -55,7 +56,7 @@ public function email(): string
         return $this->email;
     }
 
-    public function birthday(): ?string
+    public function birthday(): ?\DateTimeImmutable
```
lchrusciel:
```suggestion
    public function birthday(): ?DateTimeImmutable
```
as it is already imported

lchrusciel:

PR #550, opened 2019-09-21
Making phpstan level higher

mamazu:


lchrusciel:
Thanks, @mamazu! :tada:

PR #552, opened 2019-09-22
Adding more methods to the verify account

### LINKED ISSUE #541: Verify Account should be support other http methods
Currently Verify Account does only support GET which make sense when you have link from a email to it. But from an app perspective it should support to do also verify-account as POST or PUT request as this make more sense when manipulating data on the server.

mamazu:
Closes #541


lchrusciel:
Thanks, @mamazu! :tada:

PR #555, opened 2019-09-22
add validation to verify account action and update swagger.yml

CSchulz:


spec/Validator/Customer/VerificationTokenExistsValidatorSpec.php
```diff
@@ -0,0 +1,61 @@
...
+        $this->shouldHaveType(VerificationTokenExistsValidator::class);
+    }
+
+    function it_does_not_add_constraint_if_verification_token_is_empty(
+        UserRepositoryInterface $userRepository,
+        ExecutionContextInterface $executionContext
+    ): void {
+        $executionContext->addViolation(Argument::cetera())->shouldNotBeCalled();
+
+        $this->validate('', new VerificationTokenExists());
```
mamazu: You should split this up into two test cases.

spec/Validator/Customer/VerificationTokenExistsValidatorSpec.php
```diff
@@ -0,0 +1,61 @@
...
+    function it_is_initializable(): void
+    {
+        $this->shouldHaveType(VerificationTokenExistsValidator::class);
+    }
+
+    function it_does_not_add_constraint_if_verification_token_is_empty(
+        UserRepositoryInterface $userRepository,
+        ExecutionContextInterface $executionContext
+    ): void {
+        $executionContext->addViolation(Argument::cetera())->shouldNotBeCalled();
```
mamazu: In this case `Argument::any()` would be enough.

spec/Validator/Customer/VerificationTokenExistsValidatorSpec.php
```diff
@@ -0,0 +1,61 @@
...
+        $this->initialize($executionContext);
+    }
+
+    function it_is_initializable(): void
+    {
+        $this->shouldHaveType(VerificationTokenExistsValidator::class);
+    }
+
+    function it_does_not_add_constraint_if_verification_token_is_empty(
+        UserRepositoryInterface $userRepository,
```
mamazu: Unused parameter.

src/Validator/Customer/VerificationTokenExistsValidator.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\Validator\Customer;
+
+use Sylius\Component\User\Repository\UserRepositoryInterface;
```
mamazu: Please use the interface from the core.

src/Validator/Customer/VerificationTokenExistsValidator.php
```diff
@@ -0,0 +1,27 @@
...
+    private $userRepository;
+
+    public function __construct(UserRepositoryInterface $userRepository)
+    {
+        $this->userRepository = $userRepository;
+    }
+
+    public function validate($token, Constraint $constraint)
+    {
+        if (null !== $token && strlen($token) > 0 && null === $this->userRepository->findOneBy(['emailVerificationToken' => $token])) {
```
mamazu: Skip cases where the token is null because this will be handled in the `NotNull` constraint.

tests/Controller/Customer/VerifyApiTest.php
```diff
@@ -45,6 +45,38 @@ public function it_allows_to_verify_customer(): void
...
+     */
+    public function it_allows_to_verify_customer_and_returns_properly_error_code(): void
+    {
+        $response = $this->verifyAccount('token');
+        $this->assertResponse($response, 'customer/verify_account_token_not_exists', Response::HTTP_BAD_REQUEST);
+    }
+
+    private function verifyAccount(?string $token): Response
+    {
+        $token = $token !== null ? '?token=' . $token : '';
```
mamazu: Use the query parameter of the client.

src/Validator/Customer/VerificationTokenExistsValidator.php
CSchulz: This would result into multiple validation errors, is that wanted?

mamazu: Nope, that's why you should only handle "your" part of the validation (meaning the repository)

src/Validator/Customer/VerificationTokenExistsValidator.php
CSchulz: There is only this interface. Do you mean the class?

lchrusciel: We should use a core interface

tests/Controller/Customer/VerifyApiTest.php
```diff
@@ -45,6 +45,37 @@ public function it_allows_to_verify_customer(): void
...
+        $response = $this->verifyAccount('token');
+        $this->assertResponse($response, 'customer/verify_account_token_not_exists', Response::HTTP_BAD_REQUEST);
+    }
+
+    private function verifyAccount(?string $token): Response
+    {
+        $this->client->request(
+            'GET',
+            '/shop-api/verify-account',
+            [$token !== null ? '?token=' . $token : ''],
```
lchrusciel: Won't it work?
```suggestion
            $token !== null : ['token' => $token] ? [],
```

CSchulz: To be honest I don't know and the php doc wasn't helpful 🤣

lchrusciel:
Thank you, @CSchulz! :tada:

src/Validator/Customer/VerificationTokenExistsValidator.php
CSchulz: I don't know which one you mean. The `UserRepositoryInterface` from the user component is the only one, I have found.

lchrusciel: Ok, my bad.

PR #557, opened 2019-09-22
Made tranlation keys more specific

mamazu:


lchrusciel:
Thank you, @mamazu! :tada:

PR #562, opened 2019-09-22
WIP: country route

adrianmarte:
New "country" route to read all available countries including the provinces


lchrusciel:
Thanks a lot for the contribution! 

Nevertheless, it will be a part of 1.1 release at least. And also, I would love to provide a direct guide and set up my mind how to handle CRUD / Command<>View based endpoints

mamazu:
Instead of giving the user a list of all countries this might make sense to be part of the cart api where you can get a list of countries where the cart can be shipped.

mamazu:
This merge request can be closed since we merged #722 which is a cleaner implementation of it.

Issue #49, opened 2017-04-13
[RFC] [Feature Request] Products GRID


lchrusciel:
Currently, the implementation does not allow to sort paginated products page or reduce an amount of records. We need to add a grid to it.

CSchulz:
Querying the products without any taxonomy or similar would be also a nice addtion.

PR #564, opened 2019-09-22
add simple product name and description search

CSchulz:
Simple implementation for a product name and description search.

I don't like the always full blown output for a search f.e. if you have a search bar with result overlay, you don't need all details.

### DIFF
--- a/doc/swagger.yml
+++ b/doc/swagger.yml
@@ -553,6 +553,35 @@ paths:
                     schema:
                         $ref: "#/definitions/ProductDetails"
 
+    /products/by-phrase/{phrase}:
+        get:
+            tags:
+                - "products"
+            summary: "Show a product catalog."
+            description: "This endpoint will return a paginated list of products for given phrase."
+            operationId: "productCatalogByPhrase"
+            parameters:
+                -   name: "phrase"
+                    in: "path"
+                    description: "Phrase to search for."
+                    required: true
+                    type: "string"
+                -   name: "locale"
+                    in: "query"
+                    description: "Locale in which products should be shown."
+                    required: false
+                    type: "string"
+                -   name: "includeDescription"
+                    in: "query"
+                    description: "Should the description included in search."
+                    required: false
+                    type: "boolean"
+            responses:
+                200:
+                    description: "Show a product with the given code."
+                    schema:
+                        $ref: "#/definitions/ProductDetails"
+
     /products/by-slug/{slug}/reviews:
         parameters:
             -   name: "slug"
--- a/tests/Controller/Product/ShowCatalogByNameApiTest.php
+++ b/tests/Controller/Product/ShowCatalogByNameApiTest.php
@@ -1,0 +1,76 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Tests\Sylius\ShopApiPlugin\Controller\Product;
+
+use Symfony\Component\HttpFoundation\Response;
+use Tests\Sylius\ShopApiPlugin\Controller\JsonApiTestCase;
+
+final class ShowCatalogByNameApiTest extends JsonApiTestCase
+{
+    /**
+     * @test
+     */
+    public function it_shows_paginated_products_for_phrase(): void
+    {
+        $this->loadFixturesFromFiles(['shop.yml']);
+
+        $this->client->request('GET', '/shop-api/products/by-phrase/mug', [], [], self::CONTENT_TYPE_HEADER);
+        $response = $this->client->getResponse();
+
+        $this->assertResponse($response, 'product/product_list_page_by_phrase_mug_response', Response::HTTP_OK);
+    }
+
+    /**
+     * @test
+     */
+    public function it_shows_paginated_products_for_phrase_including_description(): void
+    {
+        $this->loadFixturesFromFiles(['shop.yml']);
+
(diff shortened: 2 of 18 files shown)


doc/swagger.yml
```diff
@@ -553,6 +553,35 @@ paths:
...
+                    description: "Locale in which products should be shown."
+                    required: false
+                    type: "string"
+                -   name: "includeDescription"
+                    in: "query"
+                    description: "Should the description included in search."
+                    required: false
+                    type: "boolean"
+            responses:
+                200:
```
mamazu: And a 404 if the channel is not found.

mamazu (approved):

tests/Controller/Product/ShowCatalogByNameApiTest.php
```diff
@@ -0,0 +1,76 @@
...
+declare(strict_types=1);
+
+namespace Tests\Sylius\ShopApiPlugin\Controller\Product;
+
+use Symfony\Component\HttpFoundation\Response;
+use Tests\Sylius\ShopApiPlugin\Controller\JsonApiTestCase;
+
+final class ShowCatalogByNameApiTest extends JsonApiTestCase
+{
+    /**
```
mamazu: Maybe also add a test for
* channel not found
* When no product is matching the search

mamazu (approved):

doc/swagger.yml
CSchulz: I am not sure how to trigger a ChannelNotFoundException, because the `channelContext` is a `CachedPerRequestChannelContext`, which uses the `CompositeChannelContext`.

It seems that the `ChannelNotFoundException` can only occur if no channels are defined at all.

If the locale does not match you also get a 404.

CSchulz:
I would like also to integrate a grid (#49) but I have no clue how to connect it with an api action.

doc/swagger.yml
mamazu: The channel is determined with the hostname. So if you provide a hostname that doesn't match anything it will return a 404.

CSchulz: I understand what you want, but I don't know exactly how to do it. Are you expecting to send the hostname header?

Furthermore I have searched the existing tests and found `ShowDetailsBySlugApiTest.it_throws_a_not_found_exception_if_channel_has_not_been_found` which is doing the same thing.

I am wondering if extracting such kind of error handling as generic approach for the shop api would be suitable. If some api needs another kind of handling f.e. `ChannelNotFoundException` the developer has to implement it on its own.

mamazu: Well there are multiple ways of handling the channel resolution in the shop API. Since the ShowDetailsBySlug also tests for the channel not found case you might as well do it for consistency.

mamazu:
> I would like also to integrate a grid (#49) but I have no clue how to connect it with an api action.

It should work out of the box if configured correctly. Have a look at how the Admin Api handles it.

doc/swagger.yml
CSchulz: I have created an issue for it #566 .

mezzat11:
@mamazu when  it will be merged ?

lchrusciel:
It will be merged after 1.0 release (what means ~ the end of the next month)

mathieudureisseix:
Hi,
Do you know when the feature will be merge ?
Best regards
Mathieu

diimpp:
If it will be merged, then it will be part of BC promise, which is not optimal as it really should be implemented via filters system, rather than separate endpoint.

Great PR though. :+1:

mamazu:
Yeah this makes much more sense. We just need to copy over the route config from the admin bundle and we are basically done.

diimpp:
@mamazu 

Personally I was thinking of ShopApiPlugin/Request filters system, where only goal is to convert querystring arguments into Command/Query variables. A sort of mini-library to handle querystring.

Here is pseudo-code to illustrate the idea.
```php
class ShowProductsRequest
{
    use FiltersTrait;

    /** @var string */
   private $channelCode;

    /** @var string */
   private $localeCode;

   public __construct(Request $request)
   {
        // Should be handled via channelContext/localeContext and channel/locale aware CommandProvider/QueryProvider.
        $this->channelCode = $request->get('channel');        
        $this->localeCode = $request->get('locale');

        $this->filters->addDefinition(new BoolFilter($publicPropertyName, $privatePropertyName, $required, $whatever));
        $this->filters->addDefinition(new SearchFilter('search', ['name', 'description']));

        $this->filters->process($request);

        // At this point ShowProductsRequest->filters should be validateable as regular request property. It should be possible to provide set of predefined filter validators.
    }

    public function getQuery(): QueryInterface
    {
        return new ShowProductsQuery($this->channelCode, $this->localeCode, $this->filters->toArray())
    }
}
```

At ViewRepository/QueryHandler/CommandHandler it should be possible to write an answering part, that will understand `$filters` format and will do automatic filtering via QueryBuillder.

So I think something along those lines will keep plugin true to CQS idea and at the same time will take some manual load from programmer.

mamazu:
Building another filter based action (despite the fact that we have all of the features in the resource bundle) doesn't really make sense, in my opinion. If we'd build something like that we should definitely only build an adapter for the ResourceProvider in the resource bundle. And we already have the endpoints that are generated by the resource bundle (eg. the address book).

Personalizing the output of it is indeed a problem. The projects that I use the resource bundle in we replaced the ViewFactory to also accept serialization data, which I think is the best solution.

PR #568, opened 2019-09-25
[Customer] Fix customer update action

lchrusciel:


lchrusciel:
Huge thank you to @matik955, who reported this issue directly to us, as it could be a security vulnerability.
giphy

mamazu:
For those of you woundering what the bug was. That the customer could update the email address and impersonate other users with it.

PR #573, opened 2019-10-03
Extending customer details tests

mamazu:
This PR fixes/changes the folloing things:
* (DX) In the tests the customer details test the login request is replaced with the trait that does the request
* (DX) Removed credentials from "customer is not logged in" test.
* (DX) Extracted requesting the endpoint into a function (and now with globally defined `CONTENT_TYPE_HEADER`)
* Fixing type bug. If you instanciate the `UpdateCustomer` with null for `subscribedToNewsletter` then the getter for it will crash as null is not part of the return type.

### DIFF
--- a/src/Command/Customer/UpdateCustomer.php
+++ b/src/Command/Customer/UpdateCustomer.php
@@ -45,7 +45,7 @@ class UpdateCustomer implements CommandInterface
         $this->birthday = $birthday;
         $this->gender = $gender;
         $this->phoneNumber = $phoneNumber;
-        $this->subscribedToNewsletter = $subscribedToNewsletter;
+        $this->subscribedToNewsletter = $subscribedToNewsletter ?? false;
     }
 
     public function firstName(): string
--- a/tests/Controller/Customer/UpdateCustomerApiTest.php
+++ b/tests/Controller/Customer/UpdateCustomerApiTest.php
@@ -53,6 +53,40 @@ JSON;
         Assert::assertEquals($customer->isSubscribedToNewsletter(), true);
     }
 
+    public function it_updates_customer_when_newsletter_is_not_set(): void
+    {
+        $this->loadFixturesFromFiles(['channel.yml', 'customer.yml']);
+        $this->logInUser('oliver@queen.com', '123password');
+
+        /** @var CustomerRepositoryInterface $customerRepository */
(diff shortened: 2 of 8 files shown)


tests/Responses/Expected/customer/update_customer.json
```diff
@@ -1,9 +1,9 @@
 {
-    "id": @integer@,
+    "id": "@integer@",
```
lchrusciel: There will be much more issues like that in our tests

mamazu: There should be no problem with regex search and replace but I would like to extract that into a different merge requst.

src/Command/Customer/UpdateCustomer.php
```diff
@@ -45,7 +45,7 @@ public function __construct(
         $this->birthday = $birthday;
         $this->gender = $gender;
         $this->phoneNumber = $phoneNumber;
-        $this->subscribedToNewsletter = $subscribedToNewsletter;
+        $this->subscribedToNewsletter = $subscribedToNewsletter ?? false;
```
lchrusciel: Can we add some tests for this case?

tests/Controller/Customer/UpdateCustomerApiTest.php
```diff
@@ -53,6 +53,40 @@ public function it_updates_customer(): void
...
+            "gender": "m",
+            "phoneNumber": "0918972132",
+        }
+JSON;
+        $this->client->request('PUT', '/shop-api/me', [], [], self::CONTENT_TYPE_HEADER, $data);
+        $response = $this->client->getResponse();
+        $this->assertResponse($response, 'customer/update_customer', Response::HTTP_OK);
+
+        /** @var CustomerInterface $customer */
+        $customer = $customerRepository->findOneByEmail('oliver@queen.com');
```
lchrusciel: This could be skipped, I suppose, it is already check in the response above :)

lchrusciel:

tests/Controller/Customer/UpdateCustomerApiTest.php
mamazu: Yeah a response code check would have been enough. But now it's already merged. :D

PR #575, opened 2019-10-04
Add missing assertation for ProductDetailsViewRepository

diimpp:


mamazu:
Please also rebase this pull request. One thing I am not really sure about if this should still be part of version 1.0 what do you think @lchrusciel

diimpp:
@mamazu by-code method has this check, but by-slug doesn't. This PR is a bug-fix, rather than feature.

https://github.com/Sylius/ShopApiPlugin/blob/8aae72292caf6a696f9a334459abd115cd914f2a/src/ViewRepository/Product/ProductDetailsViewRepository.php#L62

diimpp:
A bit off topic, @mamazu @lchrusciel there is number of commands at `src/Command/Cart`, that should be at `src/Command/Checkout` (Which doesn't exists now) and there is no locale context at this plugin. Some Queries are using request directly, so maybe plugin needs LocaleAware command provider as well.

mamazu:
1. Yes, there should be a separation between cart and checkout in the command and request classes. I didn't really consider that when refactoring the structure. 
2. If you are talking about extracting the locale from the request you are right. Because determining the locale might also be more complicated if validation (is the locale allowed in the current channel) comes to mind.

lchrusciel:
Hello Dimitri,

thanks a lot for suggestions. You are totally right with your points. About the second, we should provide support for LocaleContext (similar to ChannelContext), so we can support locales in headers for example (#582)

In terms of this PR, this check is not needed as we are sure, that this product has this channel because it is part of database query
https://github.com/Sylius/ShopApiPlugin/blob/0b3f67f6f7ae8212aef76ae662fbcd2110d3d202/src/ViewRepository/Product/ProductDetailsViewRepository.php#L46-L48

Does this explanation resolve your issue?

PR #578, opened 2019-10-07
Fix token generator service for customer's password reset handler

diimpp:


mamazu:
Could you please rebase it to the current master.

lchrusciel:
Can we write some tests for it, so we can avoid regression in the future?

mamazu:
How would you test something like this?

lchrusciel:
I'm just wondering. Maybe check in the database if the field is set?

lchrusciel:
Thanks, Dmitri! :tada:

lchrusciel:
Anyway, it was added to https://github.com/Sylius/ShopApiPlugin/releases/tag/v1.0.0-rc.3

Enjoy! :)

PR #579, opened 2019-10-09
Fixing dependencies

mamazu:
In the most recent version of phpstan's assert implementation the "implements" assertion has been added this howerver does not work with the current usage. I already made a pull request to fix it but other than that we need to exclude this version until it's fixed.


lchrusciel:

PR #585, opened 2019-10-18
[Product] Fix showing attributes for a product

GSadee:
Replaces #542

### DIFF
--- a/src/Factory/Product/ProductAttributeValueViewFactory.php
+++ b/src/Factory/Product/ProductAttributeValueViewFactory.php
@@ -41,7 +41,7 @@ final class ProductAttributeValueViewFactory implements ProductAttributeValueVie
         return $productAttributeValueView;
     }
 
-    private function resolveSelectAttribute(ProductAttributeValueInterface $productAttributeValue, $locale)
+    private function resolveSelectAttribute(ProductAttributeValueInterface $productAttributeValue, $locale): array
     {
         $values = [];
         $configuration = $productAttributeValue->getAttribute()->getConfiguration();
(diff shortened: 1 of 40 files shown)


src/Factory/Product/ProductAttributeValueViewFactory.php
```diff
@@ -18,39 +19,45 @@ public function __construct(string $productAttributeValueViewClass)
...
         /** @var ProductAttributeTranslationInterface $productAttributeTranslation */
-        $productAttributeTranslation = $productAttribute->getTranslation($locale);
+        $productAttributeTranslation = $productAttribute->getTranslation($localeCode);
         $productAttributeValueView->name = $productAttributeTranslation->getName();
 
         return $productAttributeValueView;
     }
 
-    private function resolveSelectAttribute(ProductAttributeValueInterface $productAttributeValue)
+    private function getAttributeValue(ProductAttributeValueInterface $productAttributeValue, string $localeCode)
```
mamazu: Can we type-hint this to an array?

lchrusciel: It may be string or array

mamazu: Ah okay so this method handles both cases? Choices and scalar values?

lchrusciel: Yes, we have too. One of the attribute types (multiple select) may return more than one value, therefore we will support mixin here. In the future, one can add JSON object for instance.

mamazu: Okay understandable.

lchrusciel:
Closing in favor of #586

Issue #567, opened 2019-09-25
Register handlers only for the bus they are meant for


TiMESPLiNTER:
Currently all handlers are registered for all buses. (Even the one's created by myself.) It would be great to limit the handlers to only the bus they are meant for.

See following debugging output of messenger:

```
Messenger
=========

sylius_shop_api_plugin.command_bus
----------------------------------

 The following messages can be dispatched:

 ---------------------------------------------------------------------------------------------------                       
  Sylius\ShopApiPlugin\Command\AddressBook\SetDefaultAddress                                         
      handled by sylius.shop_api_plugin.handler.set_default_address_handler                          
  Sylius\ShopApiPlugin\Command\Cart\AddressOrder                                                     
      handled by sylius.shop_api_plugin.handler.address_order_handler                                
  Sylius\ShopApiPlugin\Command\Cart\PickupCart                                                       
      handled by sylius.shop_api_plugin.handler.pickup_cart_handler                                  
  Sylius\ShopApiPlugin\Command\Cart\DropCart                                                         
      handled by sylius.shop_api_plugin.handler.drop_cart_handler                                    
  Sylius\ShopApiPlugin\Command\Cart\PutSimpleItemToCart                                              
      handled by sylius.shop_api_plugin.handler.put_simple_item_to_cart_handler                      
  Sylius\ShopApiPlugin\Command\Cart\PutVariantBasedConfigurableItemToCart                            
      handled by sylius.shop_api_plugin.handler.put_variant_based_configurable_item_to_cart_handler  
  Sylius\ShopApiPlugin\Command\Cart\PutOptionBasedConfigurableItemToCart                             
      handled by sylius.shop_api_plugin.handler.put_option_based_configurable_item_to_cart_handler   
  Sylius\ShopApiPlugin\Command\Cart\ChangeItemQuantity                                               
      handled by sylius.shop_api_plugin.handler.change_item_quantity_handler                         
  Sylius\ShopApiPlugin\Command\Cart\RemoveItemFromCart                                               
      handled by sylius.shop_api_plugin.handler.remove_item_from_cart_handler                        
  Sylius\ShopApiPlugin\Command\Cart\ChooseShippingMethod                                             
      handled by sylius.shop_api_plugin.handler.choose_shipping_method_handler                       
  Sylius\ShopApiPlugin\Command\Cart\ChoosePaymentMethod                                              
      handled by sylius.shop_api_plugin.handler.choose_payment_method_handler                        
[...]

PR #590, opened 2019-10-30
[Maintenance] Register handler only in proper bus

lchrusciel:


GSadee:
I guess that this PR fixes https://github.com/Sylius/ShopApiPlugin/issues/567

mamazu:
Thank you, Łukasz! :1st_place_medal:

PR #592, opened 2019-10-30
Adding accept language to locale listener

### LINKED ISSUE #582: Add Accept-Language header
Currently is possible to pass locale only via request query.
It would be nice to have possibility to send locale via Accept-Language header.

mamazu:
Fixes #582 
Now you can also set the locale based on the accept language headers.


mamazu:
Sounds like a more consistent way. I will do that.

PR #593, opened 2019-10-30
[Checkout] Send confirmation email after checkout is completed

lchrusciel:
Relates #250


spec/Command/Customer/SendOrderConfirmationSpec.php
```diff
@@ -0,0 +1,20 @@
...
+use PhpSpec\ObjectBehavior;
+
+final class SendOrderConfirmationSpec extends ObjectBehavior
+{
+    function let(): void
+    {
+        $this->beConstructedWith('ORDERTOKEN');
+    }
+
+    function it_has_order_token(): void
```
GSadee:
```suggestion
    function it_has_an_order_token(): void
```

spec/EventListener/Messenger/OrderCompletedListenerSpec.php
```diff
@@ -0,0 +1,28 @@
...
+use Symfony\Component\Messenger\MessageBusInterface;
+
+final class OrderCompletedListenerSpec extends ObjectBehavior
+{
+    function let(MessageBusInterface $bus): void
+    {
+        $this->beConstructedWith($bus);
+    }
+
+    function it_informs_customer_about_order_completiotion(MessageBusInterface $bus): void
```
GSadee:
```suggestion
    function it_informs_customer_about_order_completion(MessageBusInterface $bus): void
```
?

src/Handler/Cart/CompleteOrderHandler.php
```diff
@@ -19,10 +22,17 @@ final class CompleteOrderHandler
...
     private $stateMachineFactory;
 
-    public function __construct(OrderRepositoryInterface $orderRepository, StateMachineFactory $stateMachineFactory)
-    {
+    /** @var MessageBusInterface */
+    private $messageBus;
+
+    public function __construct(
+        OrderRepositoryInterface $orderRepository,
+        StateMachineFactory $stateMachineFactory,
```
GSadee:
```suggestion
        StateMachineFactoryInterface $stateMachineFactory,
```

lchrusciel:
Hey Max,

Recently I came across this issue, and I've tried to solve it the best way. As order confirmation email is placed in ShopBundle, it is better not to use it in ShopAPI plugin, as these two are not compatible atm. Triggering the event is the best way to solve it. I don't want to block anything anymore, so I don't want to refactor everything, but IMHO it makes sense to introduce this concept here and migrate to others if needed. 

I've updated my PR with some documentation.

Issue #598, opened 2019-11-11
CompleteOrderRequest should check for customer object


kayue:
In the `CompleteOrderAction`, the `email` field is optional, and when checkout as a guest and no email is provided, the customer object won't be assigned to the cart.

When customer object is missing it will cause problems later on `CustomerOrderAddressesSaver`

kayue:
The CompleteOrderAction doesn't perform any validation, this behavior should be change. We should perform inventory availability and order details check in this step.

https://github.com/Sylius/ShopApiPlugin/blob/master/src/Controller/Checkout/CompleteOrderAction.php#L49

mamazu:
The thing with the email address on checkout is a little more complicated. There has been a merge request to fix this behaviour as far as I know.
https://github.com/Sylius/ShopApiPlugin/pull/365#issue-242013792
That was the issue.

And about the availability, yes this should absolutely be validated.

kayue:
That PR was merged a year ago, so I believe it doesn't cover my scenario, which is:

- User is not logged in and customer does NOT exists

The fix can be simple, just make `email` field a required field if customer object does not exist in the complete order step. Merging #599 would be the first step.

mamazu:
The thing should be fixed and covered by tests. The reason the email address is not required is the following. If the customer is logged in then the email is taken from the logged in user (and therefore not required). The issue we faced was that a customer could try to checkout as guest and then in the last step fill in the email of an existing customer and order in their name (which is obviously a huge security issue). See this test: https://github.com/Sylius/ShopApiPlugin/blob/c476d1d2a7584c41eaffdc0e4b8c537a5491586b/tests/Controller/Checkout/CompleteOrderApiTest.php#L215

For your case we indeed don't have a test but when given the email then the `AssignCustomerToCartHandler` should handle the creation of the user as the `CustomerProvider` does it.

kayue:
I understand everything you said and I know why it wasn't required. 

You are correct when email is given then everything is fine. We just need to handle the case where no email is given, a proper error message should return. Thank you for your time.

PR #599, opened 2019-11-11
Validate complete order request

kayue:
The first step to fix https://github.com/Sylius/ShopApiPlugin/issues/598 is to perform validation in `CompleteOrderAction`

### DIFF
--- a/spec/Validator/Cart/CartReadyForCheckoutValidatorSpec.php
+++ b/spec/Validator/Cart/CartReadyForCheckoutValidatorSpec.php
@@ -1,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+namespace spec\Sylius\ShopApiPlugin\Validator\Cart;
+
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+use SM\Factory\FactoryInterface;
+use SM\StateMachine\StateMachineInterface;
+use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+use Sylius\ShopApiPlugin\Validator\Constraints\CartReadyForCheckout;
+use Symfony\Component\Validator\Context\ExecutionContextInterface;
+
+class CartReadyForCheckoutValidatorSpec extends ObjectBehavior
+{
+    function let(
+        RepositoryInterface $repository,
+        FactoryInterface $stateMachineFactory,
+        ExecutionContextInterface $context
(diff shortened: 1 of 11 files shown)


mamazu:
What does it validate at the moment, if you haven't have any validation specified.

kayue:
@mamazu Ideally it should re-validate inventory availability, which can be implemented later. 

Nevertheless, Sylius is meant to be extendable, even the ShopAPI doesn't have any validation specified, it should still perform validation to allower developer to add custom validator.

kayue:
Also due to inconsistent implementation in action classes, error is being return in multiple different formats. 

For example, invalid cart token, it returns proper error 400 in `PutItemToCartAction`:

```
{
  "code": 400,
  "message": "Validation failed",
  "errors": {
    "token": [
      "Cart does not exist."
    ]
  }
}
```

However, in `CompleteOrderAction` it just return error 500 due to lack of validation:

```
{
  "code": 500,
  "message": "Order with ca5ac343-f09d-4cf0-8d34-374d77cd2ca91 token has not been found."
}
```

In `SummarizeAction` it returns error 404

```
{
  "code": 404,
  "message": "Cart with given id does not exists"
}
```

mamazu:
Currently as you found the order complete doesn't validate anything and therefore only asserts that the cart is there (which then will cause an error => 500) We should definitely add the same validators as the other endpoints have as well.

lchrusciel:
Hey @kayue,

are you willing to add test for this case?

spec/Validator/Cart/CartReadyForCheckoutValidatorSpec.php
```diff
@@ -0,0 +1,66 @@
...
+
+        $this->validate('CART_TOKEN', new CartReadyForCheckout());
+    }
+
+    function it_adds_a_violation_if_cart_can_not_checkout(
+        RepositoryInterface $repository,
+        OrderInterface $order,
+        FactoryInterface $stateMachineFactory,
+        StateMachineInterface $stateMachine,
+    ExecutionContextInterface $context
```
lchrusciel:
```suggestion
        ExecutionContextInterface $context
```

lchrusciel:
Thanks, Ka! :1st_place_medal:

kayue:
Thanks @lchrusciel for adding the tests. I didn't know what to test but now I learned.

lchrusciel:
All credits to @mamazu. I just asked him for help ;)

Issue #506, opened 2019-07-31
Replace Province name with province code.


mamazu:
In the address mapper and therefore in all address related things the province code is optional. When not providing a province code the tax calculation and zoning might not be correctly calculated. So I would suggest to remove the province name as a field from the address and only rely on the province code.

For those shops that don't rely on provinces, they would just create one default province that can be attached to all requests.

alexander-schranz:
Think we should do this before the 1.0 release else zones work not correctly.

mamazu:
It already does. If you are giving it a code it will take that provinceCode into account. But if you don't do that or you just give it the name of a province then it will not do anything.

AndreasA:
Actually it mostly works already. Sylius by default uses province code, if the country has provinces and that seems mostly fine with the Shop API.
**HOWEVER:** The `\Sylius\ShopApiPlugin\Factory\AddressBook\AddressViewFactory` only serilizes the provinceName and not the code and therefore, the provinceCode will not be in the JSON data array.

PR #600, opened 2019-11-11
[ENHANCEMENT] Added province code to AddressView

AndreasA:
Releates: #506 506

### DIFF
--- a/spec/Factory/AddressBook/AddressViewFactorySpec.php
+++ b/spec/Factory/AddressBook/AddressViewFactorySpec.php
@@ -29,6 +29,7 @@ final class AddressViewFactorySpec extends ObjectBehavior
         $address->getCountryCode()->willReturn('GB');
         $address->getCity()->willReturn('London');
         $address->getPostcode()->willReturn('NMW');
+        $address->getProvinceCode()->willReturn('GB-NWP');
         $address->getProvinceName()->willReturn('Greater London');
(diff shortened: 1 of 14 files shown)


mamazu:
First of all, thanks for your pull request. This is definitely an improvement. Currently we have two ways of handleing the address, one part is the view factory that you changed which affects how the addresses are rendered in the cart. The other part is the CRUD bundle, that renders the addresses with it's own logic (controlled by the serializer).

It would be nice if you could fix the tests and maybe also add one that checks that the cart now shows the correct address fields.

AndreasA:
Sorry for not fixing the tests first but I had to exist my train. I will probably fix them tomorrow morning and add another test.

In regards to CRUD etc. from what I can tell everywhere else the provinceCode seems to work fine (when using my JS code with it). The only issue was the AddressView.

One thing though that is a bit `annoying` when saving an address (address book) you can only send either provinceCode (if the country has provinces assigned to it) or provinceName if the country does not have provinces assigned to it. Means unnecessary data changes when creating generalized code but that isn't that big an issue.

AndreasA:
@mamazu Fixed the tests and added a test for address with province codes instead of province names.

mamazu:
Thanks, this pull request looks great. :+1:

mamazu:
Thank you, Andreas! :1st_place_medal:

Issue #589, opened 2019-10-26
[BUG] On order complete the assign customer to cart triggers the order processor - Which can change previously shown prices


AndreasA:
Hi,

I just found out that the complete action triggers the assign customer to cart action which in theory is correct (https://github.com/Sylius/ShopApiPlugin/blob/master/src/Controller/Checkout/CompleteOrderAction.php#L46). 
However, that Action also triggers the order processor: https://github.com/Sylius/ShopApiPlugin/blob/master/src/Handler/Cart/AssignCustomerToCartHandler.php#L46

As the customer was shown all the valid prices in the summary, doing so can be really bad as it could potentially change the submitted prices and finish the checkout of the order.

AndreasA:
Another reason why this isn't a good ideas:
- A customer gets reduced prices
- Now, someone else knows that customer's e-mail address and does a guest checkout with their e-mail but their shipping destination and even billing destination, they would then get their reduced prices which doesn't make sense.

PR #603, opened 2019-11-23
Check order integrity on checkout complete

Zales0123:
Fixes https://github.com/Sylius/ShopApiPlugin/issues/589

The main idea right now is to check the total of the order before and after customer assignment (and therefore - cart recalculation). This is the same thing as we do in [Sylius itself](https://github.com/Sylius/Sylius/blob/27c5b4970ccc086393ce0d643b0065e87cab065c/src/Sylius/Bundle/ShopBundle/EventListener/OrderIntegrityChecker.php#L78). The next step would be also to check the promotion integrity 🐎 

#SymfonyHackday

### DIFF
--- a/src/Controller/Checkout/CompleteOrderAction.php
+++ b/src/Controller/Checkout/CompleteOrderAction.php
@@ -8,6 +8,7 @@ use FOS\RestBundle\View\View;
 use FOS\RestBundle\View\ViewHandlerInterface;
 use Sylius\ShopApiPlugin\Command\CommandInterface;
 use Sylius\ShopApiPlugin\CommandProvider\CommandProviderInterface;
+use Sylius\ShopApiPlugin\Exception\OrderTotalIntegrityException;
 use Sylius\ShopApiPlugin\Exception\WrongUserException;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -57,6 +58,15 @@ final class CompleteOrderAction
                     View::create(
                         'You need to be logged in with the same user that wants to complete the order',
                         Response::HTTP_UNAUTHORIZED
+                    )
+                );
+            }
+
+            if ($previousException instanceof OrderTotalIntegrityException) {
+                return $this->viewHandler->handle(
+                    View::create(
+                        ['code' => Response::HTTP_BAD_REQUEST, 'message' => $exception->getMessage()],
+                        Response::HTTP_BAD_REQUEST
                     )
                 );
             }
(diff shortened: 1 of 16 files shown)


src/Controller/Checkout/CompleteOrderAction.php
```diff
@@ -59,9 +62,27 @@ public function __invoke(Request $request): Response
...
+
             throw $exception;
         }
 
         return $this->viewHandler->handle(View::create(null, Response::HTTP_NO_CONTENT));
     }
+
+    private function provideCorrectCompleteOrderCommand(Request $request): CommandInterface
+    {
+        if (null !== $request->request->get('email')) {
```
mamazu: Does this also account for guest checkout and impersonation checkout? So that I can just enter your email address and then do a guest checkout and Sylius adds the order in your account?

src/Controller/Checkout/CompleteOrderAction.php
```diff
@@ -59,9 +62,27 @@ public function __invoke(Request $request): Response
                 );
             }
 
+            if ($previousException instanceof OrderTotalIntegrityException) {
```
mamazu: Can't we use a validator and then just handle the response of the validator instead of building something on our own?

src/Processor/SafeOrderProcessor.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        $oldTotal = $order->getTotal();
+
+        $this->orderProcessor->process($order);
+
+        if ($order->getTotal() !== $oldTotal) {
+            throw new OrderTotalIntegrityException();
+        }
+
+        $this->orderManager->flush();
```
lchrusciel: We need a test for the case when the cart was recalculated and customer would like to finish a checkout anyway.

src/Controller/Checkout/CompleteOrderAction.php
lchrusciel: I would vote for validation here as well. But #599 may be needed to solve it properly.

src/Controller/Checkout/CompleteOrderAction.php
lchrusciel: What do you think about placing this resolution logic to the provider itself?

mamazu: That would be a good idea.

src/Controller/Checkout/CompleteOrderAction.php
mamazu: Working on it.

Zales0123:
Ok, so we've just had a (not-so) quick talk with @lchrusciel about this implementation and there is a new vision 😄 

1. Deprecate setting an email in CompleteOrder action
2. Create a new action for assigning the customer to the order 
3. Validate the order after completing it for total integrity and customer assignment

And it should still work for current implementations and allow us to control this integrity changes during the last order recalculation. Sounds good? 🔉

PR #606, opened 2019-11-23
Adding a test for skipped shipping and payment.

### LINKED ISSUE #597: Be able to skip payment
Actualy, if I apply a coupon that makes the cart to be at cost 0, payment is removed, witch is correct. But the issue is, there is no way to complete the checkout neither is there an endpoint to skip payment.

mamazu:
Closes #597


mamazu:
@lchrusciel Could you have a look at this. I can't reproduce the error the CI is giving me.

PR #611, opened 2019-12-11
[Doc] fix `checkoutState` parameter in `PlacedOrder` swagger definition

EmilMassey:
I think the intention here was to display order's state (then the enum would be correct), but in fact checkout state is returned as in `Cart` definition.

Note that the checkout state will always be `completed` for placed orders.

### DIFF
--- a/doc/swagger.yml
+++ b/doc/swagger.yml
@@ -1758,15 +1758,16 @@ definitions:
             checkoutState:
                 description: "Current state of a checkout."
                 type: "string"
-                default: "cart"
+                default: "checkout"
                 externalDocs:
                     description: "Find out more about checkout states in the Sylius documentation."
                     url: "http://docs.sylius.com/en/latest/book/orders/checkout.html#checkout-state-machine"
                 enum:
                     - "cart"
-                    - "new"
-                    - "fulfilled"
-                    - "cancelled"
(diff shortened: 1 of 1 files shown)


EmilMassey:
I think displaying order's state would be useful for some, though. Do you think we should add it to the view? I could work on implementation.

doc/swagger.yml
```diff
@@ -1758,15 +1758,16 @@ definitions:
             checkoutState:
                 description: "Current state of a checkout."
                 type: "string"
-                default: "cart"
+                default: "checkout"
```
mamazu: The default value is not part of the possible enum values?

EmilMassey: Thank you for pointing it out! It was obviously a mistake, the state is `completed`.

mamazu:
The default state is probably `cart` for carts and only `completed` for orders.

doc/swagger.yml
```diff
@@ -1758,15 +1758,16 @@ definitions:
             checkoutState:
                 description: "Current state of a checkout."
                 type: "string"
-                default: "cart"
+                default: "completed"
```
lchrusciel:
```suggestion
                default: "cart"
```

Issue #569, opened 2019-09-26
Bad response when trying to add address


matik955:
When trying to add address not beeing logged response is: 
```
"code":500,"message":"Argument 1 passed to Sylius\\Component\\Core\\Factory\\AddressFactory::createForCustomer() must implement interface Sylius\\Component\\Core\\Model\\CustomerInterface, null given 
```

It should return status code 401 for example.

mamazu:
This is probably due to #462

lchrusciel:
These endpoints should be hidden behind the firewall. They don't have any sense without it.

mamazu:
True.

Amr3zzat:
I think Address book in only should be behind the firewall in the routes true?

I have just added  access control rule to fix this issue.
in security.yaml
`   access_control:`
`        - { path: "%sylius_shop_api.security.regex%/address-book", role: ROLE_USER}
`

mamazu:
Yes.

Amr3zzat:
@mamazu It should be added to theme readme? Should I create PR with it or there is a better fix?

mamazu:
The best way to fix this is indeed in the routes to protect them in the `security.yml`. Theoretically this should also include the profile routes (eg. /me) This should also be documented in the swagger file which is partly is.

mamazu:
But if you could create a merge request to fix this issue this would be great.

PR #616, opened 2019-12-27
Protect Address Book and me routes

Amr3zzat:
Protect Address book and me routes in security.yaml 
related issues #569

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,10 @@ The latest documentation is available [here](https://app.swaggerhub.com/apis/Syl
                 guard:
                     authenticators:
                         - lexik_jwt_authentication.jwt_token_authenticator
+       access_control:
+       - { path: "%sylius_shop_api.security.regex%/address-book", role: ROLE_USER}
+       - { path: "%sylius_shop_api.security.regex%/me", role: ROLE_USER}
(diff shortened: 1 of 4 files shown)


mamazu:
Thank you for your pull request. One thing I was woundering when looking through the documentation is: In the docs it says that the endpoint returns a 500 if the user does not own the address. If this behaviour has also changed then could you please update the documentation and add a note in the `UPGRADE.md` that this changed.

Amr3zzat:
@mamazu Thanks for review , I have updated the docs and the upgrade.

mamazu:
Could you please rebase your pull request to the current master and also add the authorization documentation to the new "change password route"?

Amr3zzat:
@mamazu Sure I will do that

Amr3zzat:
@mamazu  I have rebased the master , I checked the change password route in docs , the new protection will match with new route docs , It will return 401

mamazu:
I don't really get why the address book routes return a 404 when the customer does not have access to those routes instead of a 403 but as this seems to be Sylius default I am okay with it.

mamazu:
Thanks, Amr! :tada:

Issue #614, opened 2019-12-27
[BUG] Gender Field needs validation update user details


Amr3zzat:
When Sending gender field in "Updates currently logged in users details." End Point ,
It returns 500 when sending more than one character , It should return 400, It needs validation on gender field length.

<img width="798" alt="Screen Shot 2019-12-27 at 3 53 41 PM" src="https://user-images.githubusercontent.com/31932378/71519636-125b6480-28c1-11ea-8c1f-e0eaf2391957.png">

Yasser-G:
also it need to be clear what are acceptable letters that represents `gender` value

Amr3zzat:
@Yasser-G  I did that in my PR

mamazu:
Merged.

PR #617, opened 2019-12-27
Add Validation on Customer Gender

Amr3zzat:
Add Validation on Gender field [Update Customer Request]
Gender Filed will only accept these letters ['u', 'm', 'f'].
related issue #614


mamazu:
Looks good, so far. If you could fix the tests it would be great. And maybe add a test that tests the validation (so that when the user enters an invalid gender it returns a 400).

Amr3zzat:
@mamazu I have fixed the old test , add new test for this case

mamazu:
Thank you, Amr! :tada:

Issue #618, opened 2020-01-01
Request password reset needs validation


Amr3zzat:
There are two scenarios:
First:
Enter empty email 
Expected:
it returns 400 with validation error 
Actual:
it returns 500
<img width="735" alt="Screen Shot 2020-01-01 at 9 13 18 AM" src="https://user-images.githubusercontent.com/31932378/71638999-f49e4e80-2c76-11ea-9ee4-a0812bbf6c29.png">

 Second:
Enter email not exists in users
Expected:
it returns 400 with validation error 
Actual:
it returns 500
<img width="647" alt="Screen Shot 2020-01-01 at 9 12 42 AM" src="https://user-images.githubusercontent.com/31932378/71638995-df292480-2c76-11ea-8e0d-f9fea7965d46.png">

Amr3zzat:
I was working in this issue

I have fixed the first part of the issue in my PR

About the second part “User not found”
I think it should return 204 in all cases if the user entered a valid email even it doesn’t exists for security considerations, true?

PR #621, opened 2020-01-10
Add email validation to RequestResetPassword

Amr3zzat:
Add email Validation to RequestResetPassword
#618

### DIFF
--- a/spec/Handler/Customer/SendResetPasswordTokenHandlerSpec.php
+++ b/spec/Handler/Customer/SendResetPasswordTokenHandlerSpec.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace spec\Sylius\ShopApiPlugin\Handler\Customer;
 
+use InvalidArgumentException;
 use PhpSpec\ObjectBehavior;
 use Sylius\Component\Core\Model\ShopUserInterface;
 use Sylius\Component\Mailer\Sender\SenderInterface;
@@ -12,7 +13,6 @@ use Sylius\ShopApiPlugin\Command\Customer\SendResetPasswordToken;
 use Sylius\ShopApiPlugin\Exception\UserNotFoundException;
 use Sylius\ShopApiPlugin\Handler\Customer\SendResetPasswordTokenHandler;
 use Sylius\ShopApiPlugin\Mailer\Emails;
-use InvalidArgumentException;
 
 final class SendResetPasswordTokenHandlerSpec extends ObjectBehavior
 {
--- a/spec/Handler/Customer/SendResetPasswordTokenHandlerSpec.php
+++ b/spec/Handler/Customer/SendResetPasswordTokenHandlerSpec.php
@@ -5,13 +5,13 @@ declare(strict_types=1);
 namespace spec\Sylius\ShopApiPlugin\Handler\Customer;
 
 use PhpSpec\ObjectBehavior;
-use Prophecy\Argument;
 use Sylius\Component\Core\Model\ShopUserInterface;
 use Sylius\Component\Mailer\Sender\SenderInterface;
 use Sylius\Component\User\Repository\UserRepositoryInterface;
 use Sylius\ShopApiPlugin\Command\Customer\SendResetPasswordToken;
 use Sylius\ShopApiPlugin\Handler\Customer\SendResetPasswordTokenHandler;
 use Sylius\ShopApiPlugin\Mailer\Emails;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
 final class SendResetPasswordTokenHandlerSpec extends ObjectBehavior
 {
@@ -45,15 +45,13 @@ final class SendResetPasswordTokenHandlerSpec extends ObjectBehavior
         $userRepository->findOneByEmail('example@customer.com')->willReturn($user);
         $user->getPasswordResetToken()->willReturn(null);
 
-        $this->shouldThrow(\InvalidArgumentException::class)->during('__invoke', [new SendResetPasswordToken('example@customer.com', 'WEB_GB')]);
+        $this->shouldThrow(NotFoundHttpException::class)->during('__invoke', [new SendResetPasswordToken('example@customer.com', 'WEB_GB')]);
     }
 
-    function it_continues_if_user_not_found(
-        UserRepositoryInterface $userRepository,
-        SenderInterface $sender,
-        ShopUserInterface $user
+    function it_throws_an_exception_if_user_has_not_been_found(
+        UserRepositoryInterface $userRepository
     ): void {
-        $userRepository->findOneByEmail('amr@amr.com')->willReturn(null);
-        $sender->send(Argument::any())->shouldNotBeCalled();
+        $userRepository->findOneByEmail('example@customer.com')->willReturn(null);
+        $this->shouldThrow(NotFoundHttpException::class)->during('__invoke', [new SendResetPasswordToken('example@customer.com', 'WEB_GB')]);
     }
 }
--- a/spec/Handler/Customer/SendResetPasswordTokenHandlerSpec.php
+++ b/spec/Handler/Customer/SendResetPasswordTokenHandlerSpec.php
@@ -46,4 +46,13 @@ final class SendResetPasswordTokenHandlerSpec extends ObjectBehavior
 
         $this->shouldThrow(\InvalidArgumentException::class)->during('__invoke', [new SendResetPasswordToken('example@customer.com', 'WEB_GB')]);
     }
+
+    function it_continues_if_user_not_found(
+        UserRepositoryInterface $userRepository,
+        SenderInterface $sender,
+        ShopUserInterface $user
+    ): void {
+        $userRepository->findOneByEmail('amr@amr.com')->willReturn(null);
+        $sender->send(Emails::EMAIL_RESET_PASSWORD_TOKEN, ['example@customer.com'], ['user' => $user, 'channelCode' => 'WEB_GB'])->shouldNotBeCalled();
+    }
 }
--- a/spec/Handler/Customer/SendResetPasswordTokenHandlerSpec.php
+++ b/spec/Handler/Customer/SendResetPasswordTokenHandlerSpec.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace spec\Sylius\ShopApiPlugin\Handler\Customer;
 
 use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
 use Sylius\Component\Core\Model\ShopUserInterface;
 use Sylius\Component\Mailer\Sender\SenderInterface;
 use Sylius\Component\User\Repository\UserRepositoryInterface;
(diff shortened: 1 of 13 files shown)


mamazu:
I have fixed the codestyle in your branch and added the information about the change in behavior in the change logs and rebased it onto the master so the pipeline should be green now.

spec/Handler/Customer/SendResetPasswordTokenHandlerSpec.php
```diff
@@ -37,14 +37,6 @@ function it_handles_emailing_user_with_verification_email(
         $this(new SendResetPasswordToken('example@customer.com', 'WEB_GB'));
     }
 
-    function it_throws_an_exception_if_user_has_not_been_found(
```
mamazu: We should add a test that checks that the class does nothing if the user was not found.

src/Handler/Customer/SendResetPasswordTokenHandler.php
```diff
@@ -31,14 +31,13 @@ public function __invoke(SendResetPasswordToken $resendResetPasswordToken): void
...
-
-        Assert::notNull($user, sprintf('User with %s email has not been found.', $email));
-        Assert::notNull($user->getPasswordResetToken(), sprintf('User with %s email has not verification token defined.', $email));
-
-        $this->sender->send(
-            Emails::EMAIL_RESET_PASSWORD_TOKEN,
-            [$email],
-            ['user' => $user, 'channelCode' => $resendResetPasswordToken->channelCode()]
-        );
+        if (null !== $user) {
```
mamazu: An early return would make the logic clearer here I think.

Amr3zzat:
@mamazu Thanks for your review I will check them, I have updated my local branch, I got stuck for some time, but It worked thanks

src/Handler/Customer/SendResetPasswordTokenHandler.php
Amr3zzat: Check Now

spec/Handler/Customer/SendResetPasswordTokenHandlerSpec.php
Amr3zzat: Done

spec/Handler/Customer/SendResetPasswordTokenHandlerSpec.php
```diff
@@ -54,4 +46,13 @@ function it_throws_an_exception_if_user_has_not_verification_token(
...
         $this->shouldThrow(\InvalidArgumentException::class)->during('__invoke', [new SendResetPasswordToken('example@customer.com', 'WEB_GB')]);
     }
+
+    function it_continues_if_user_not_found(
+        UserRepositoryInterface $userRepository,
+        SenderInterface $sender,
+        ShopUserInterface $user
+    ): void {
+        $userRepository->findOneByEmail('amr@amr.com')->willReturn(null);
+        $sender->send(Emails::EMAIL_RESET_PASSWORD_TOKEN, ['example@customer.com'], ['user' => $user, 'channelCode' => 'WEB_GB'])->shouldNotBeCalled();
```
mamazu: Here we should assert that the method was not called with any parameters: `$sender->send(Arguments::any())->shouldNotBeCalled();`

src/Controller/Customer/RequestPasswordResettingAction.php
```diff
@@ -34,12 +38,14 @@ final class RequestPasswordResettingAction
     public function __construct(
         ViewHandlerInterface $viewHandler,
         MessageBusInterface $bus,
+        ValidationErrorViewFactoryInterface $validationErrorViewFactory,
```
lchrusciel: Due to BC we should add it as the last argument + add notice about depreciation.

Something like this:
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/AdminBundle/Controller/ImpersonateUserController.php#L50-L52

src/Handler/Customer/GenerateResetPasswordTokenHandler.php
```diff
@@ -30,10 +29,11 @@ public function __invoke(GenerateResetPasswordToken $generateResetPasswordToken)
 
         /** @var ShopUserInterface $user */
         $user = $this->userRepository->findOneByEmail($email);
-
-        Assert::notNull($user, sprintf('User with %s email has not been found.', $email));
```
lchrusciel: The handler should throw an exception anyway, I suppose.

src/Handler/Customer/SendResetPasswordTokenHandler.php
```diff
@@ -31,14 +31,15 @@ public function __invoke(SendResetPasswordToken $resendResetPasswordToken): void
 
         /** @var ShopUserInterface $user */
         $user = $this->userRepository->findOneByEmail($email);
-
-        Assert::notNull($user, sprintf('User with %s email has not been found.', $email));
```
lchrusciel: The handler should throw an exception anyway, I suppose.

src/Handler/Customer/GenerateResetPasswordTokenHandler.php
mamazu: Where should this be caught? In the controller?

lchrusciel: IMHO, we shouldn't catch them, because on application level it is a mistake. Validation of request should protect us from this scenario anyway

mamazu: Yeah, makes sense.

Amr3zzat: I didn't get it! my logic is that the handler won't do any thing if the user email not found like sylius behaviour, The endpoint will return 204
`if (null === $user) {
            return;
        }`
Why It should throw exception?

mamazu: True, but the this file is there for handling a command. This command COULD be called from the console where an error message might be helpful. So I agree with Lukasz that throwing an exception makes sense but you should definitely catch that and ignore it in the controller.

Amr3zzat: Okay I got it, I will add the commit

src/Handler/Customer/SendResetPasswordTokenHandler.php
Amr3zzat: Check Now

src/Handler/Customer/SendResetPasswordTokenHandler.php
```diff
@@ -31,10 +31,12 @@ public function __invoke(SendResetPasswordToken $resendResetPasswordToken): void
...
-        Assert::notNull($user->getPasswordResetToken(), sprintf('User with %s email has not verification token defined.', $email));
-
+        if (null === $user) {
+            throw new NotFoundHttpException('User with given email does not exist!');
```
lchrusciel: Can you throw `InvalidArgumentExcpetion` (or custom exception), as handlers should be HTTP agnostic?

By design, handlers should be infrastructure agnostic because we don't know where they came from, and the most important thing for us is to keep the application in proper state.

Please, at least revert to the previous implementation

src/Controller/Customer/RequestPasswordResettingAction.php
```diff
@@ -36,22 +41,38 @@ public function __construct(
...
+                Response::HTTP_BAD_REQUEST
+            ));
+        }
 
-        $this->bus->dispatch($this->generateResetPasswordTokenCommandProvider->getCommand($request));
-        $this->bus->dispatch($this->sendResetPasswordTokenCommandProvider->getCommand($request, $channel));
+        try {
+            $this->bus->dispatch($this->generateResetPasswordTokenCommandProvider->getCommand($request));
+            $this->bus->dispatch($this->sendResetPasswordTokenCommandProvider->getCommand($request, $channel));
+        } catch (HandlerFailedException $notFoundHttpException) {
```
lchrusciel: Can you remove try-catch block? Validation should protect us from accessing this part of code. While muting exceptions may be really misleading for users.

mamazu: We agreed to handle it so that the user of the API does not get an error message if the email of the resetting user does not exist. Because then a potential attacker would know that this email address is registered in this shop.

lchrusciel: It makes some sense. Can we leave the comment then, so no one will be surprised with it?

src/Handler/Customer/SendResetPasswordTokenHandler.php
Amr3zzat: I created a new Custom Exception please check now

src/Controller/Customer/RequestPasswordResettingAction.php
Amr3zzat: I think I made the logic clearer in the new commit, Please review

src/Controller/Customer/RequestPasswordResettingAction.php
```diff
@@ -43,7 +44,8 @@ public function __construct(
         CommandProviderInterface $generateResetPasswordTokenCommandProvider,
         ChannelBasedCommandProviderInterface $sendResetPasswordTokenCommandProvider,
         ?ValidationErrorViewFactoryInterface $validationErrorViewFactory
-    ) {
+    )
```
lchrusciel:
```suggestion
    ) {
```

src/Controller/Customer/RequestPasswordResettingAction.php
```diff
@@ -43,7 +44,8 @@ public function __construct(
...
+    {
```
lchrusciel:
```suggestion
```

src/Exception/UserNotFoundException.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\Exception;
+
+
+class UserNotFoundException extends \InvalidArgumentException
```
lchrusciel:
```suggestion
final class UserNotFoundException extends \InvalidArgumentException
```

src/Exception/UserNotFoundException.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\Exception;
+
+
```
lchrusciel:
```suggestion
```

src/Exception/UserNotFoundException.php
```diff
@@ -0,0 +1,20 @@
...
+    public static function occur(): self
+    {
+        return new self('User with given email does not exist!');
+    }
+
+    public static function withEmail(string $email): self
+    {
+        return new self(sprintf('User with email %s has not been found.', $email));
+    }
+
```
lchrusciel:
```suggestion
```

Amr3zzat:
@mamazu  @lchrusciel  Thanks for your review, it was so helpful, I have fixed the specs.

mamazu:
If you can fix the tests then it should be good to go.

src/Handler/Customer/GenerateResetPasswordTokenHandler.php
```diff
@@ -30,9 +30,9 @@ public function __invoke(GenerateResetPasswordToken $generateResetPasswordToken)
 
         /** @var ShopUserInterface $user */
```
mamazu: This can also return null. 
```php
 /** @var ShopUserInterface|null $user */
```

Amr3zzat:
@mamazu Many Thanks for your help, I fixed them

mamazu:
Thanks, Amr! :1st_place_medal:

PR #623, opened 2020-01-15
Fixing php type errors instead of ignoring them

mamazu:
With the upgrade of phpstan to the newer version some typehints were not fixed but only ignored for simplicity. This pull request fixes them and adds documentation to the right places.

One thing I need to mention though is that most of those can be replaced with real typehints in php like `string|int` if we can assume that the id's in Sylius are integers (which they are for the most part).

Another thing is that the variable naming of the payment and shipping selector commands is a little confusing. Because the command needs an integer (index) but the identifier could also be a string (payment or shipment-code). This needs to be renamed to clarify. But this would be a BC break, so maybe for a later time.


src/Factory/Customer/CustomerViewFactoryInterface.php
```diff
@@ -5,8 +5,9 @@
 namespace Sylius\ShopApiPlugin\Factory\Customer;
 
 use Sylius\Component\Core\Model\CustomerInterface;
+use Sylius\ShopApiPlugin\View\Customer\CustomerView;
 
 interface CustomerViewFactoryInterface
 {
-    public function create(CustomerInterface $customer);
+    public function create(CustomerInterface $customer): CustomerView;
```
lchrusciel: BC Break :(  even if it makes sense

src/Validator/Address/AddressExistsValidator.php
```diff
@@ -28,20 +28,25 @@ public function __construct(
     }
 
     /**
+     * @param string|int               $id
```
lchrusciel:
```suggestion
     * @param string|int $id
```

src/Factory/Customer/CustomerViewFactoryInterface.php
mamazu: So not part of 1.1?

lchrusciel: not really. rather some note in docs?

mamazu: What we should do is add this as a doc block so phpstan is happy and the code still runs the same but I would definitely somehow mark this behavior to return something other then a `CustomerView` as deprecated.

lchrusciel:
Thanks, @mamazu! :tada:

PR #624, opened 2020-01-15
Removing dead validation for nonexisting class

mamazu:


lchrusciel:
Thanks, @mamazu! :tada:

Issue #241, opened 2017-11-18
[BUG Report] checkout error


sdleiw:
Hi, 

I could not make `PUT /checkout/{cartid}/address` call after updating my local api server, got a 500 error

Type error: Return value of Sylius\\Bundle\\CoreBundle\\Checkout\\CheckoutResolver::getRequestedGraph() must be of the type string, null returned"

have noticed that in v1.0.3 update the checkout_resolve config was removed, but it was suggested  for the shop-api server in the readme of this repo, can someone take a look at that, by the way, with this config the put address API call works

my current api server dependencies
- sylius v1.0.3
- syliusShopApiPlugin dev-master
- nelmio/cors-bundle v1.5.3

Hailong:
Hi @sdleiw , I saw the same error when loading the checkout page, it might be the same issue for API. Changing the path matching pattern will resolve the problem, you could check it out from my PR here: https://github.com/Sylius/SyliusShopApiPlugin/pull/246

bochkarev-artem:
The same issue with version sylius 1.6.1 and sylius/shop-api-plugin v1.0.0-rc.2
The documentation of [this mr](https://github.com/Sylius/ShopApiPlugin/pull/246) helped
Although works also with this configuration:
`sylius_shop:
    checkout_resolver:
        pattern: "%sylius.security.shop_regex%/checkout/.+"`

bochkarev-artem:
i would sugget reopenning the issue, or updating documentation

misnard:
> The same issue with version sylius 1.6.1 and sylius/shop-api-plugin v1.0.0-rc.2
> The documentation of [this mr](https://github.com/Sylius/ShopApiPlugin/pull/246) helped
> Although works also with this configuration:
> 
> ```yaml
> sylius_shop:
>     checkout_resolver:
>         pattern: "%sylius.security.shop_regex%/checkout/"
> ```

I found this config inside of this file config/packages/_sylius.yaml

Thanks.

mamazu:
Is this still an issue in the current configuration? Because now there is no config key like `checkout_resolver` in the `README` anymore.

misnard:
Yes, i face this problem and i didn't found any help in README hopefully this topic help me with some research.

mamazu:
This seems to be still an issue.

mamazu:
This seems to be still an issue.

alexander-schranz:
Can confirm that it still exist.

[...]

PR #625, opened 2020-01-22
Delete incompatibility message as it isnt the case

remy-theroux:
Ass discuss in the slack of Sylius and as i see in many places, ShopApi is compatible with the ShopBundle so it's better to delete this confusing message.
I am new to Sylius and i take time to search why it's not compatible. It would be helpfull for others to know there is no problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,7 @@
 
 The latest documentation is available [here](https://app.swaggerhub.com/apis/Sylius/sylius-shop-api/1.0.0). If you are looking for more information how the system works have a look at the [cookbook](doc/Cookbook.md)
 
-## Installation
-
-##### Before installing SyliusShopApiPlugin, you should disable all SyliusShopBundle's dependencies. You cannot use these packages together.
+## Installation.
 
 1. Run `composer require sylius/shop-api-plugin` and, when asked if you want to execute the Flex recipe, answer 'Yes'.
 2. Extend config files:


mamazu:
In general I agree, we have solved the problem that was responsible for putting the error message into the documentation. However there is still no guarantee that everything will work perfectly fine (we might have not discovered it yet). See: https://github.com/Sylius/ShopApiPlugin/issues/241

The next problem is that in our test suite we only test the shopapi functions (with the shop as a dependency). What we should really do is also check if the shop is working (aka running the test suit from the `sylius/sylius` package).

remy-theroux:
Ok i will try to find time to automate that if it's a convenient solution for you. I am very new to Sylius.

mamazu:
Sure you can try to do that. Testing the integration would be a great thing to have.

The thing is no matter how you want to word it. ShopApi and the shop use different mechanics and different routes for the same thing. (eg. with the checkout) So there will be incompatibilities. And therefore it is currently **not recommended to use the shop and the api together**. If the compatibility issues are huge or small is a different question. You can try to work around them but this is at your own risk.

BTW: If the checkout process is buggy, which in my opionion is the most important part of an online shop, then it would be justified to call it "not compatible" :D

GSadee:
I am aware of one problem that this record is still needed for. This plugin overwrites the service (`CartBlamerListener`) by which the cart in Shop is not assigned to the customer (in the database) after logging in. After solving that issue, we will probably be able to remove that message I guess.

mamazu:
I'd vote to close this pull request as it is still not compatible.

remy-theroux:
I close it, i have not enough experience in Sylius so you are probably right.

PR #626, opened 2020-01-22
More comprehensive error messages for checkout validation

### LINKED ISSUE #574: More verbose error messages on denied state changes during checkout
Currently if I want to complete an order and I haven't gone through defining a payment method all I get when calling `PUT /checkout/{cartToken}/complete` is:

```
{
    "code": 500,
    "message": "Order with {cartToken} token cannot be completed."
}
```

it would be very helpful to have a more verbose error message and maybe another code. As `500` somehow implies that there is a server error and somewhere you should be able to find a stacktrace or at least log entry.

A message like:

> Order with {cartToken} token cannot be completed. Please set a payment method first.

would be way clearer.

mamazu:
Closes #574

### DIFF
--- a/src/Validator/Cart/CartReadyForCheckoutValidator.php
+++ b/src/Validator/Cart/CartReadyForCheckoutValidator.php
@@ -5,6 +5,8 @@ declare(strict_types=1);
 namespace Sylius\ShopApiPlugin\Validator\Cart;
 
 use SM\Factory\FactoryInterface;
+use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Core\OrderCheckoutStates;
 use Sylius\Component\Core\OrderCheckoutTransitions;
 use Sylius\Component\Resource\Repository\RepositoryInterface;
 use Sylius\ShopApiPlugin\Validator\Constraints\CartReadyForCheckout;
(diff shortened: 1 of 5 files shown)


src/Validator/Cart/CartReadyForCheckoutValidator.php
```diff
@@ -28,16 +30,36 @@ public function __construct(RepositoryInterface $cartRepository, FactoryInterfac
...
-        if (!$stateMachine->can(OrderCheckoutTransitions::TRANSITION_COMPLETE)) {
-            /** @var CartReadyForCheckout $constraint */
-            $this->context->addViolation($constraint->message);
+        if ($stateMachine->can(OrderCheckoutTransitions::TRANSITION_COMPLETE)) {
+            return;
         }
+
+        /** @var CartReadyForCheckout $constraint */
+        $validationMessage = $constraint->message;
+        switch ($cart->getState()) {
```
lchrusciel: Wouldn't it be more granular if we would just create 4 different validators?

mamazu: Yes, but then you'd have to have control over in which order they are validated. It doesn't make sense to return a response like payment method not chosen if the cart hasn't been addressed yet. What we could do however is make multiple constraints that are validated by the same validator. Because the logic of the validator is always the same. I am not so sure however if that is a good idea.

lchrusciel:

PR #628, opened 2020-01-29
Use Channel Context instead of getChannel

### LINKED ISSUE #577: Multiple "getChannel"s ignore "enabled" property of Channel
I've encountered two cases, but codebase needs to be grepped for more.

https://github.com/Sylius/ShopApiPlugin/blob/master/src/ViewRepository/Product/ProductCatalogViewRepository.php#L88
https://github.com/Sylius/ShopApiPlugin/blob/master/src/ViewRepository/Product/ProductLatestViewRepository.php#L61

Amr3zzat:
Use Channel Context Directly in ViewRepository instead of passing channel code and use getChannel

### DIFF
--- a/src/Controller/Product/ShowLatestProductAction.php
+++ b/src/Controller/Product/ShowLatestProductAction.php
@@ -22,8 +22,7 @@ final class ShowLatestProductAction
     public function __construct(
         ViewHandlerInterface $viewHandler,
         ProductLatestViewRepositoryInterface $productLatestQuery
-    )
-    {
+    ) {
         $this->viewHandler = $viewHandler;
         $this->productLatestQuery = $productLatestQuery;
     }
@@ -31,7 +30,6 @@ final class ShowLatestProductAction
     public function __invoke(Request $request): Response
     {
         try {
-
             return $this->viewHandler->handle(View::create($this->productLatestQuery->getLatestProducts(
                 $request->query->get('locale'),
                 $request->query->getInt('limit', 4)
--- a/src/Controller/Product/ShowLatestProductAction.php
+++ b/src/Controller/Product/ShowLatestProductAction.php
@@ -6,8 +6,6 @@ namespace Sylius\ShopApiPlugin\Controller\Product;
 
 use FOS\RestBundle\View\View;
 use FOS\RestBundle\View\ViewHandlerInterface;
-use Sylius\Component\Channel\Context\ChannelContextInterface;
-use Sylius\Component\Channel\Context\ChannelNotFoundException;
 use Sylius\ShopApiPlugin\ViewRepository\Product\ProductLatestViewRepositoryInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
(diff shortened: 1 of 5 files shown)


src/ViewRepository/Product/ProductLatestViewRepository.php
```diff
@@ -26,21 +25,31 @@ final class ProductLatestViewRepository implements ProductLatestViewRepositoryIn
...
-    public function getLatestProducts(string $channelCode, ?string $localeCode, int $count): ProductListView
+    public function getLatestProducts(?string $localeCode, int $count): ProductListView
     {
-        $channel = $this->getChannel($channelCode);
+        try {
+            /** @var ChannelInterface $channel */
+            $channel = $this->channelContext->getChannel();
+
+        } catch (ChannelNotFoundException $exception) {
+            throw new NotFoundHttpException('Channel has not been found.');
```
mamazu: The implementation in the repository should not be bound to HTTP.

mamazu:
@lchrusciel What do you think? I don't know if this has any wider implications.

src/Controller/Product/ShowLatestProductAction.php
```diff
@@ -21,31 +19,21 @@ final class ShowLatestProductAction
...
-
     public function __construct(
         ViewHandlerInterface $viewHandler,
-        ProductLatestViewRepositoryInterface $productLatestQuery,
-        ChannelContextInterface $channelContext
+        ProductLatestViewRepositoryInterface $productLatestQuery
     ) {
         $this->viewHandler = $viewHandler;
         $this->productLatestQuery = $productLatestQuery;
-        $this->channelContext = $channelContext;
```
lchrusciel: Even if it's not a BC berak, I would leave this argument here, so we will not make a bc break in the future and trigger deprecation error, that this class should not be injected anymore and will be removed in ShopAPI 2.0

src/ViewRepository/Product/ProductLatestViewRepository.php
```diff
@@ -26,21 +25,29 @@ final class ProductLatestViewRepository implements ProductLatestViewRepositoryIn
     /** @var SupportedLocaleProviderInterface */
     private $supportedLocaleProvider;
 
+    /** @var ChannelContextInterface */
+    private $channelContext;
+
     public function __construct(
-        ChannelRepositoryInterface $channelRepository,
```
lchrusciel: 🚨 bc break 🚨

src/ViewRepository/Product/ProductLatestViewRepositoryInterface.php
```diff
@@ -8,5 +8,5 @@
 
 interface ProductLatestViewRepositoryInterface
 {
-    public function getLatestProducts(string $channelCode, ?string $localeCode, int $count): ProductListView;
+    public function getLatestProducts(?string $localeCode, int $count): ProductListView;
```
lchrusciel: 🚨BC BREAK 🚨

lchrusciel (commented):
bc breaks cannot be merged to master (unless we are planing version 2.0). 

Also, @Amr3zzat can you elaborate a little bit more, how it will fix the issue? It seems to be rather a refactor, with exactly the same behavior for me.

Amr3zzat:
@lchrusciel Yes it is the same behaviour,  it is just refactor to get the channel directly in the ViewRepository instead of passing the code from the action, I removed the issue mentioned to fix that.

lchrusciel:
Hey Amr, thanks a lot for your work, but as I said, due to bc breaks in this PR, I cannot merge it.

I hope you understand.

PR #630, opened 2020-01-31
docs: in PlacedOrder definition add number property

EmilMassey:


mamazu:
Thank you, Emil! :tada:

lchrusciel:
Should we update the swagger file now then? It's on dev-master, so we need to keep the version in sync, I suppose 

ref. https://github.com/Sylius/ShopApiPlugin/blob/master/doc/swagger.yml#L4

mamazu:
True. I will have an eye on that in the future.

PR #633, opened 2020-02-17
Add documentation details on how to extends views

remy-theroux:
This detail would have helped me if it were here


remy-theroux:
I did not see any CONTRIBUTE markdown file so i create a simple PR.

lchrusciel:
Thank you, Rémy! :1st_place_medal:

PR #635, opened 2020-02-18
Update AddProductReviewByCodeRequest.php

aike-ps:
Fix bug. Rating must be only Int.

### DIFF
--- a/src/Request/Product/AddProductReviewByCodeRequest.php
+++ b/src/Request/Product/AddProductReviewByCodeRequest.php
@@ -34,7 +34,7 @@ class AddProductReviewByCodeRequest implements ChannelBasedRequestInterface
     {
         $this->code = $request->attributes->get('code');
         $this->title = $request->request->get('title');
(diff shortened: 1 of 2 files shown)


mamazu:
Could you please merge this together with the other merge request you did.

mamazu:
One thing I am thinking about is that you cahnged the default value. With `get` the default value if the property does not exist is `null` and the default value of `getInt` is `0` so there is no way of differenticating between not rated and rated very badly.

aike-ps:
now not working at all, good luck )))

aike-ps:
```
AddProductReviewByCode
 public function __construct(
        string $productCode,
        string $channelCode,
        string $title,
        **int $rating,**
        string $comment,
        string $email
    )
```

mamazu:
Thanks for pointing that out. Yes the store rating hast to be an integer but the fact that field is filled with something other than null should be validated at the request level. So if you create a review without giving it a rating then the validator should return a 400 and never trigger the command.

PR #636, opened 2020-02-23
Adding a documentation validator to the project

mamazu:
I added some use statements to the cookbook so that people know which classes are the right ones.

This validates 
* the XML (only syntax)
* the fact that the code is valid (linting)
* the use statements that I added


lchrusciel:
Thank you, @mamazu! :tada:

Issue #637, opened 2020-02-26
Cart response doesn't contain total without discount


antonioperic:
itemsTotal contains discounted price, I assume this should show totalPrice without discount, as we have discounted price in total?

```
{
    "tokenValue": "25898c3d-2278-4b1c-8468-3f045b8950d2",
    "channel": "HR_SPLIT",
    "currency": "EUR",
    "locale": "en_US",
    "checkoutState": "cart",
    "items": {
        "ffe5dff6-6a5b-37ba-b7ad-695b9d60b06a-11.06.2020": {
            "code": "ffe5dff6-6a5b-37ba-b7ad-695b9d60b06a",
            "name": "Atraction ticket \"nihil\" bluecave",
            "image": {
                "code": "header",
                "path": "75/73/10483ddd2f14d4615ec641a3342b.jpeg",
                "cachedPath": "https://api-dev.litto.co/media/cache/sylius_shop_api/75/73/10483ddd2f14d4615ec641a3342b.jpeg"
            },
            "date": "11.06.2020",
            **"total": 183,**
            "variants": [
                {
                    "cartItemId": 6,
                    "code": "ffe5dff6-6a5b-37ba-b7ad-695b9d60b06a-variant-1",
                    "name": "Adults 11.06.2020",
                    "type": "Adults",
                    **"total": 183,
                    "quantity": 14,**
                    "date": "11.06.2020"
                }
            ]
        }
    },
    **"totals": {
        "total": 183,
        "items": 183,
        "taxes": 37,
        "shipping": 0,
        "promotion": -8959
    },**
    "payments": [
        {
            "state": "cart",
            "method": {
                "code": "bank_transfer",
                "name": "Bank transfer",
                "description": "Recusandae et quae corporis rerum et.",
                "instructions": ""
            },
            "price": {
                "current": 183,
                "currency": "EUR"
            }
        }
    ],
    "shipments": [],
    "cartDiscounts": {
        "christmas": {
            "name": "Christmas",
            "amount": {
                "current": -8959,
                "currency": "EUR"
            }
        }
    }
}
```

antonioperic:
possible my mistake

lchrusciel:
Can you provide more details about your configuration? It seems, that some cart discounts maybe should be placed on item level

antonioperic:
@lchrusciel 
simple one : https://take.ms/0acdBL

mamazu:
[...]

PR #638, opened 2020-02-26
Fixed showing price total for items in cart response

antonioperic:


antonioperic:
fixes #637

mamazu:
I just added a BC Break, as it is modifying a property and changing its value.

PR #643, opened 2020-04-09
Fixing the documentation parser

mamazu:
After the update of the documentation parser the executable is renamed.


lchrusciel:
Thanks, @mamazu! :tada:

PR #644, opened 2020-04-15
Allow customizing picked up cart's locale

dnna:
Currently when picking up a new cart, the default channel locale is always used.

This PR fixes this by enabling the use of an optional ?locale URL parameter to customize the cart's locale.

### DIFF
--- a/src/CommandProvider/ChannelBasedCommandProvider.php
+++ b/src/CommandProvider/ChannelBasedCommandProvider.php
@@ -5,7 +5,9 @@ declare(strict_types=1);
 namespace Sylius\ShopApiPlugin\CommandProvider;
 
 use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\Component\Locale\Context\LocaleContextInterface;
 use Sylius\ShopApiPlugin\Command\CommandInterface;
+use Sylius\ShopApiPlugin\Command\LocaleAwareCommandInterface;
 use Sylius\ShopApiPlugin\Request\ChannelBasedRequestInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Validator\ConstraintViolationListInterface;
@@ -20,12 +22,16 @@ final class ChannelBasedCommandProvider implements ChannelBasedCommandProviderIn
     /** @var ValidatorInterface */
     private $validator;
 
-    public function __construct(string $requestClass, ValidatorInterface $validator)
+    /** @var LocaleContextInterface|null */
+    private $localeContext;
+
+    public function __construct(string $requestClass, ValidatorInterface $validator, ?LocaleContextInterface $localeContext = null)
     {
         Assert::implementsInterface($requestClass, ChannelBasedRequestInterface::class);
 
         $this->requestClass = $requestClass;
         $this->validator = $validator;
+        $this->localeContext = $localeContext;
     }
 
     public function validate(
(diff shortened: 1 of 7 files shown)


dnna:
@lchrusciel I'm happy to make the change, but note there are multiple instances where `$request->query->get('locale')` is already used in the shop API.

Indicatively check the following:
- https://github.com/Sylius/ShopApiPlugin/blob/master/src/Controller/Product/ShowProductDetailsBySlugAction.php#L45
- https://github.com/Sylius/ShopApiPlugin/blob/master/src/Controller/Product/ShowProductDetailsByCodeAction.php#L45
- https://github.com/Sylius/ShopApiPlugin/blob/master/src/Controller/Product/ShowLatestProductAction.php#L44
- https://github.com/Sylius/ShopApiPlugin/blob/master/src/Controller/Product/ShowProductReviewsBySlugAction.php#L44
- https://github.com/Sylius/ShopApiPlugin/blob/master/src/Controller/Product/ShowProductCatalogByTaxonSlugAction.php#L47
- https://github.com/Sylius/ShopApiPlugin/blob/master/src/Controller/Product/ShowProductCatalogByTaxonCodeAction.php#L47

dnna:
I have made the requested changes so that `LocaleContext` is used instead of request `query`. To make this work, I had to create a new `ChannelAndLocaleBasedRequestInterface` and a `ChannelAndLocaleBasedCommandProvider`.

For `ChannelAndLocaleBasedRequestInterface` I extended the original `ChannelBasedRequestInterface` instead of having a completely separate interface in order to minimize BC breaks.

mamazu:
Yeah localization is indeed still a problem in the ShopApi. But this looks good. I would prefer to move it in to the main provider and then add an interface that only specifies the `setLocale` function. And then in the provider if the interface is implemented call the function. What do you think?

dnna:
That approach makes a lot of sense. I'll try to implement it that way and get back.

dnna:
@mamazu I have reimplemented it the way you mentioned. I didn't use Symfony's `LocaleAwareInterface` because I was afraid it might be somehow integrated with Symfony and get unexpectedly triggered by the container or request listeners, so I created a new `LocaleAwareCommandInterface`.

I implemented it in a way that allows localeContext to be nullable to avoid BC breaks if someone has decorated the provider or extended the command. I can't see a way that it might break but if you do please let me know.

mamazu:
Looks good. This could be then extended to all other commands.

mamazu:
Thank you, Dimosthenis! :tada:

src/CommandProvider/ChannelBasedCommandProvider.php
```diff
@@ -20,12 +22,16 @@ final class ChannelBasedCommandProvider implements ChannelBasedCommandProviderIn
...
+    /** @var LocaleContextInterface|null */
+    private $localeContext;
+
+    public function __construct(string $requestClass, ValidatorInterface $validator, ?LocaleContextInterface $localeContext = null)
     {
         Assert::implementsInterface($requestClass, ChannelBasedRequestInterface::class);
 
         $this->requestClass = $requestClass;
         $this->validator = $validator;
+        $this->localeContext = $localeContext;
```
lchrusciel, line 34: We should trigger error if it won't be defined

mamazu: You mean because of backwards compatibility?

lchrusciel: Just to inform people that it will be in future required and not passing it is deprecated

mamazu: But shouldn't this be a ensured by the constructor?

mamazu: Agreed.

dnna: It's not required in the context of this PR, the code is built to handle it gracefully when its null and just uses the default channel locale.

If it needs to be made required for other reasons it should trigger a deprecation instead of an error IMHO.

PR #647, opened 2020-05-19
[Checkout] defaults routing parameters

pierre-H:


mamazu:
Thank you for your pull request but I am afraid that this doesn't do anything. There is no code that is reading this route information. This only works on resource controllers which the controllers in the route aren't.

pierre-H:
Actually, without these lines, I've got this error in every checkout route :

> Notice: Trying to access array offset on value of type null

from :

https://github.com/Sylius/Sylius/blob/552c4a00e6897f7137207f83c6293c66f7f026b1/src/Sylius/Bundle/CoreBundle/Checkout/CheckoutResolver.php#L91

lchrusciel:
I would say, we should customize the configuration, and do not access these lines in the first place. Perhaps this place could be used for that?

mamazu:
Doesn't the comment in the merge request fix the problem, if you change the path of the `checkout_resolver`?

pierre-H:
> Doesn't the comment in the merge request fix the problem, if you change the path of the `checkout_resolver`?

You're right : adding the config fixes the problem.

Thank you !

PR #651, opened 2020-06-01
Adding validator for payment methods

mamazu:


spec/Validator/Cart/PaymentMethodAvailableValidatorSpec.php
```diff
@@ -0,0 +1,94 @@
...
+        $orderRepository->findOneBy(['tokenValue' => 'NON_EXISTING', 'state' => 'cart'])->willReturn($cart);
+        $cart->getPayments()->willReturn(new ArrayCollection([$payment->getWrappedObject()]));
+
+        $paymentMethodsResolver->getSupportedMethods($payment)->shouldBeCalled()->willReturn([$paymentMethod]);
+        $paymentMethod->getCode()->willReturn('paypal');
+
+        $context->buildViolation('sylius.shop_api.checkout.payment_method_not_available')->willReturn(
+            $violationBuilder
+        )
+        ;
```
lchrusciel, line 68:
```suggestion
        );
```

src/Validator/Cart/PaymentMethodAvailableValidator.php
```diff
@@ -0,0 +1,62 @@
...
+        OrderRepositoryInterface $orderRepository,
+        PaymentMethodsResolverInterface $paymentMethodsResolver
+    ) {
+        $this->orderRepository = $orderRepository;
+        $this->paymentMethodsResolver = $paymentMethodsResolver;
+    }
+
+    public function validate($value, Constraint $constraint): void
+    {
+        /** @var ChoosePaymentMethodRequest $value */
```
lchrusciel, line 35: What do you think about adding an assertion here as well?

src/Validator/Constraints/PaymentMethodExists.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\Validator\Constraints;
+
+use Sylius\ShopApiPlugin\Validator\Cart\PaymentMethodExistsValidator;
+use Symfony\Component\Validator\Constraint;
+
+final class PaymentMethodExists extends Constraint
```
lchrusciel, line 10: Shouldn't it be only property constraint?

src/Validator/Cart/PaymentMethodExistsValidator.php
```diff
@@ -0,0 +1,33 @@
...
+
+    public function __construct(PaymentMethodRepositoryInterface $paymentMethodRepository)
+    {
+        $this->paymentMethodRepository = $paymentMethodRepository;
+    }
+
+    /** {@inheritdoc} */
+    public function validate($value, Constraint $constraint): void
+    {
+        Assert::isInstanceOf($constraint, PaymentMethodExists::class);
```
lchrusciel, line 26: I'm not sure about having two different behaviors in two validators in the same PR. Can we either ensure constraint or skip it in both?

mamazu: So you mean if a payment method does not exist it is also not valid for the order? So we should validate that the order can have this payment method?

src/Validator/Constraints/PaymentMethodExists.php
mamazu: Yes, by default it is but I added the explicit `getTargets` method to the constraint now.

src/Validator/Cart/PaymentMethodAvailableValidator.php
mamazu: Yes, forgot that.

src/Validator/Cart/PaymentMethodExistsValidator.php
lchrusciel: No, just having here `Assert::isInstanceOf($constraint, PaymentMethodExists::class);` and there is nothing alike in the second validator

mamazu: Done that. Yeah, forgot that.

lchrusciel:
Thanks, @mamazu! :tada:

PR #676, opened 2020-10-08
Fixing the Checkout state validator.

### LINKED ISSUE #665: [Bug] CartReadyForCheckoutValidator checks against order state, not order checkout state
https://github.com/Sylius/ShopApiPlugin/blob/3ace12724f53c330d6b894fe2bbe639ecc140d76/src/Validator/Cart/CartReadyForCheckoutValidator.php#L47

mamazu:
Closes #665


lchrusciel (approved):
Is there a possibility to cover a new case, which will protect us from the regression?

mamazu:
I could add one but I don't know how someone can manage to get the cart into such a state.

lchrusciel:
We should try to complete cart, which is not addresses/shipment is not selected/payment is not selected

mamazu:
Valid point I will add that.

mamazu:
We already have a test for this:
https://github.com/Sylius/ShopApiPlugin/blob/master/tests/Controller/Checkout/CompleteOrderApiTest.php#L391

lchrusciel:
Thanks, @mamazu! :tada:

Issue #667, opened 2020-08-10
Cart runs in a deadlock on multiple calls


ddanninger:
Hi guys,

i have two calls running on every application load parallel
 
`/shop-api/me`
and 
`/shop-api/carts/%TOKEN%`

everything works greats. But as soon as I apply a coupon they conflict with each other. One of those will always result in a 500 Internal Server error

Error on either one of them:
`An exception occurred while executing 'UPDATE sylius_order SET items_total = ?, adjustments_total = ?, total = ?, updated_at = ? WHERE id = ?' with params [86626564, -40000000, 46626564, \"2020-08-10 21:34:10\", 298]:\n\nSQLSTATE[40001]: Serialization failure: 1213 Deadlock found when trying to get lock; try restarting transaction`

Any help or guidance would be appreciated

mamazu:
Hello, thanks for reporting this bug. The problem occours if the cart is viewed and recalculated at the same time.

In both cases it looks like you are just viewing it but actually behind the scenes it gets recalculated every time there is an authenticated request for this user (eg. the `/me` endpoint). This is a known problem. One solution would be to remove the cart recalculation logic entirely but this might men that you will get an outdated cart.

ddanninger:
@mamazu thank you for you fast response, really appreciate that. 

in my case, it seems like the cart rules are changing constantly on every request without even changing something on the cart is that how it's supposed to be because of checking the expiration dates on promotions? 

Do you think if I would create a custom action merging the /shop-api/me & /shop-api/carts/%TOKEN% it would less conflict? Or any other ideas on how I could apply a quick fix?

mamazu:
Yes, the cart is getting recalculated no matter what. It could also be that product prices have changed or taxation and so on.

Well what you could do is add the cart to the `/me` endpoint. It would muddy the waters on what endpoints are for what but this would be the easiest fix.
On the other hand you could also disable the cart recalculation logic, this would require some more digging but in the end it's the cleaner solution.

diimpp:
It's not just `/me` endpoint, we're having deadlocks in `checkout` and whatever places just from concurrent requests.

Another problem is when deadlock doesn't actually happen, but order totals are incorrect, like shipping price is not included in order total or some phantom price adjustments exists.

The only solution, that I can think of is to introduce business transaction wrap for the whole `orderProcessing`. For example with pessimistic lock via https://symfony.com/doc/4.4/components/lock.html

mamazu:
[...]

Issue #681, opened 2020-10-23
[Bug] Core UserCartRecalculationListener triggers order processing for any endpoint for authorized user, resulting in deadlocks


diimpp:
If `shop user` is logged in and have at least one previously assigned cart, then [CoreBundle/EventListener/UserCartRecalculationListener](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/EventListener/UserCartRecalculationListener.php) will trigger `orderProcessing` on **each** authorized request to **any** shop-api endpoint.

This happens due the fact that `ShopApiPlugin` doesn't define `CartContext` and `Core\CartContext` considers last assigned cart as current cart, which is not true for shop-api. It also expected be triggered only at login in html website, but in shop-api it happens in each request with JWT token.

This results in metric ton of deadlocks for any authorized user and renders SPA app unusable.

Solution is to override listener and disable it.

mamazu:
And then we should just recalculate the cart on every request inside the `shop-api/checkout/` routes right?

diimpp:
I don't think we should additionally recalculate cart in every request at cart or checkout as most of those endpoints already doing their own `order processing`.

Original sylius strategy doesn't quite fit shop-api stateless state and any new strategies will heavily depend on how frontend will use api endpoints. (For example new `/carts/{token}/refresh`, which can be called on login by frontend)

So I think best course of action is to just disable `Core\UserCartRecalculationListener` and don't do anything else.

/Offtopic,
in my opinion shop-api v1 is kind of done and not salvageable as it was released too early and only v2 with redesign will save it. (All contexts should be re-defined in shop-api, `Query` object should be introduces in addition to `Command`, etcetera)
So we should look more in maintenance, than new features.

For example I've implemented `out of stock validator` in my project, but it's not pretty and runs on modified version of shop-api and it would be plenty of work to make proper PR.

PR #682, opened 2020-10-29
Decorating the UserCartRecalculationListener

JakobTolkemit:
With this the `UserCartRecalculationListener` listens to `lexik_jwt_authentication.on_jwt_created` instead of `security.interactive_login`, recalculating the cart only on an actual login.

Fixes: https://github.com/Sylius/ShopApiPlugin/issues/667 https://github.com/Sylius/ShopApiPlugin/issues/681


src/EventListener/UserCartRecalculationListener.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Sylius\ShopApiPlugin\EventListener;
+
+use Lexik\Bundle\JWTAuthenticationBundle\Event\JWTCreatedEvent;
+use Sylius\Bundle\CoreBundle\EventListener\UserCartRecalculationListener as CoreListener;
+use Sylius\Bundle\UserBundle\Event\UserEvent;
+use Sylius\Component\User\Model\UserInterface;
+use Symfony\Component\Security\Http\Event\InteractiveLoginEvent;
+
+class UserCartRecalculationListener
```
diimpp, line 13: final

src/EventListener/UserCartRecalculationListener.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\EventListener;
+
+use Lexik\Bundle\JWTAuthenticationBundle\Event\JWTCreatedEvent;
+use Sylius\Bundle\CoreBundle\EventListener\UserCartRecalculationListener as CoreListener;
```
diimpp, line 8: CoreUserCartRecalculationListener

src/Resources/config/services/listener.xml
```diff
@@ -0,0 +1,12 @@
+<container>
```
diimpp, line 1: I think you need to import this file at `src/Resources/config/services.xml` for it to work

diimpp, line 1: Container missing regular `xlmns` definitions

diimpp, line 1: If you chose to define new config file, then you need to move existent event listener services over here as well

src/Resources/config/services/listener.xml
```diff
@@ -0,0 +1,12 @@
...
+    <services>
+        <service class="Sylius\ShopApiPlugin\EventListener\UserCartRecalculationListener"
+                 id="sylius.shop_api_plugin.event_listener.user_cart_recalculation_listener"
```
diimpp, line 4: Usually `id` should be placed first

src/Resources/config/services/listener.xml
```diff
@@ -0,0 +1,12 @@
...
+                 decorates="sylius.listener.user_cart_recalculation"
+        >
+            <argument type="service" id="sylius.shop_api_plugin.event_listener.user_cart_recalculation_listener.inner" />
+            <tag name="kernel.event_listener" event="sylius.user.security.implicit_login" method="recalculateCartWhileLogin" />
+            <tag name="kernel.event_listener" event="lexik_jwt_authentication.on_jwt_created" method="onJwtLogin" />
```
diimpp, line 9: How about `method="onJwtCreated"` to keep it in line with event name

src/EventListener/UserCartRecalculationListener.php
```diff
@@ -0,0 +1,34 @@
...
+use Sylius\Bundle\UserBundle\Event\UserEvent;
+use Sylius\Component\User\Model\UserInterface;
+use Symfony\Component\Security\Http\Event\InteractiveLoginEvent;
+
+class UserCartRecalculationListener
+{
+    /** @var CoreListener */
+    private $recalculationListener;
```
diimpp, line 16: `$decoratedUserCartRecalculationListener`

src/EventListener/UserCartRecalculationListener.php
```diff
@@ -0,0 +1,34 @@
...
+    public function recalculateCartWhileLogin(InteractiveLoginEvent $event): void
+    {
+        $this->recalculationListener->recalculateCartWhileLogin($event);
+    }
+
+    public function onJwtLogin(JWTCreatedEvent $event): void
+    {
+        /** @var UserInterface $user */
+        $user = $event->getUser();
+        $this->recalculationListener->recalculateCartWhileLogin(new UserEvent($user));
```
diimpp, line 32: This doesn't feel like a proper implementation and overuse of `UserEvent`

src/EventListener/UserCartRecalculationListener.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /** @var CoreListener */
+    private $recalculationListener;
+
+    public function __construct(CoreListener $recalculationListener)
+    {
+        $this->recalculationListener = $recalculationListener;
+    }
+
+    public function recalculateCartWhileLogin(InteractiveLoginEvent $event): void
```
diimpp, line 23: >     public function recalculateCartWhileLogin(InteractiveLoginEvent $event): void 

`InteractiveLoginEvent $event` doesn't match original `Symfony\Component\EventDispatcher\Event $event` method signature.

Does it even work with `onJwtLogin` supplying `UserEvent` (which does not implement InteractiveLoginEvent) over here?

src/Resources/config/services/listener.xml
```diff
@@ -0,0 +1,12 @@
+<container>
+    <services>
+        <service class="Sylius\ShopApiPlugin\EventListener\UserCartRecalculationListener"
+                 id="sylius.shop_api_plugin.event_listener.user_cart_recalculation_listener"
+                 decorates="sylius.listener.user_cart_recalculation"
+        >
+            <argument type="service" id="sylius.shop_api_plugin.event_listener.user_cart_recalculation_listener.inner" />
+            <tag name="kernel.event_listener" event="sylius.user.security.implicit_login" method="recalculateCartWhileLogin" />
```
diimpp, line 8: Just a question, I don't remember from my testing -- is implicit login works properly with JWT?

diimpp (changes requested):

diimpp:
This doesn't solve #667, which prime reason are concurrent frontend calls, not internal over-use of order processing.

As for #681, personally I consider it undesirable side effect and that stateless shop-api shouldn't offer this functionality at all, moreover if a proper stateless cart context would be implemented, then this feature will never be called.

Though for the sake of a bit less broken shop-api plugin it can be done this way for 1.*

src/Resources/config/services/listener.xml
JakobTolkemit: Apparently it's only dispatched in the `Sylius\Bundle\UserBundle\Security\UserLogin::login` method. This method is only called when a user registers in the ShopBundle.
So, I guess there is no reason to keep it.

src/EventListener/UserCartRecalculationListener.php
JakobTolkemit: Really, the whole reason I did this, is because the core listener expects a `Symfony\Component\EventDispatcher\Event` but the JWTCreatedEvent extends the `Symfony\Contracts\EventDispatcher\Event`.
Now I override the class instead, I think there no reason to decorate it.

src/Resources/config/services/event_listener.xml
```diff
@@ -0,0 +1,31 @@
+<container xmlns="http://symfony.com/schema/dic/services">
```
diimpp, line 1:
```xml
<container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
```

According to sylius xml definitions, it should be like this

src/Resources/config/services/event_listener.xml
```diff
@@ -0,0 +1,31 @@
...
+    <services>
```
diimpp, line 2: This file should be named `listeners.xml` according to sylius/sylius naming convention. It a sort of nitpick, but due BC promise, it should be named similar from the start.

diimpp, line 2: Missing global public declaration

```
<defaults public="true" />
```

I'm not sure why, but it should be public in accordance with definition for `Sylius\Bundle\CoreBundle\EventListener\UserCartRecalculationListener` and because services that are moved from `services.xml` are public by default and they need to continue to be public due to BC

src/Resources/config/services/event_listener.xml
```diff
@@ -0,0 +1,31 @@
...
+        <service id="sylius.shop_api_plugin.event_listener.user_registration_listener"
```
diimpp, line 3: Services `sylius.shop_api_plugin.event_listener.messenger.cart_picked_up_listener` and `sylius.shop_api_plugin.event_listener.messenger.order_completed_listener` should be moved over here as well.

diimpp (changes requested):

src/EventListener/UserCartRecalculationListener.php
diimpp: I agree, override is better

diimpp (approved):
Thank you, Jakob.

@mamazu @lchrusciel guys, if you agree with the approach, let's not marinate this issue for months :)

src/EventListener/UserCartRecalculationListener.php
```diff
@@ -0,0 +1,40 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\EventListener;
+
+use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Order\Context\CartContextInterface;
+use Sylius\Component\Order\Context\CartNotFoundException;
+use Sylius\Component\Order\Processor\OrderProcessorInterface;
+use Symfony\Contracts\EventDispatcher\Event;
```
lchrusciel, line 11:
```suggestion
use Symfony\Component\EventDispatcher\Event;
```

Or we should probably add contracts to composer

src/EventListener/UserCartRecalculationListener.php
```diff
@@ -0,0 +1,40 @@
...
+use Webmozart\Assert\Assert;
+
+final class UserCartRecalculationListener
```
lchrusciel, line 14: Is this class any different from the original implementation?

src/EventListener/UserCartRecalculationListener.php
lchrusciel: Oh, now I see it... https://github.com/Sylius/ShopApiPlugin/pull/682#discussion_r517482564 What would you say for using JWT class explicitly? 

Also, can we maybe make it explicit, that we are removing service from the container (for example in CompilerPass with link to issue on Github) and declare separate service? Or at least link in the comment that is right now, to current issues on Github?

mamazu (approved):

mamazu:
Thanks, Jakob! :1st_place_medal:

Issue #683, opened 2020-10-30
Prevent checking out an empty cart


TiMESPLiNTER:
Currently it's possible to check out an empty cart. This doesn't make sense and is not possible via the storefront from what I know. The goal would be to streamline this behavior between the storefront and the API and check whether there is at least one item in cart before allowing to proceed with the checkout.

mamazu:
It would be very easy to add a validator for this. The question is where should this validation show? When the user is about to complete the order? Or when the user tries to address the cart?

TiMESPLiNTER:
Good question. I would say at very beginning of the checkout process so when the user tries to address the cart then.

diimpp:
Wasn't it already solved at `checkout/complete` api endpoint? Additional validator checks should be executed on each `checkout/*` endpoint due that items can be removed from cart via `carts/delete` endpoint at any time, leaving checkout in un-completable state

mamazu:
Hm, nope. I have just checked the `CartReadyForCheckoutValidator` only checks the state of the order not the order amount.

The problem with validating it every time throughout the checkout process is that this is (in my opinion) too much work. A cart is either valid or invalid. And addressing, payment selection and shipping won't change it. I mean if a customer tried to checkout with an empty cart and then only notices it on the last step, at least he tried. :D
But the main issue for having one validation call at the start of the checkout is that the checkout process is not very closed. Because at any time the customer can remove items from the cart.

Another option would be to add the call to a delete call on the item (which I would think is a very rare call).

mamazu:
This should be done in the merge request as far as I can tell right?

PR #684, opened 2020-11-06
Add validation for disabled product or product variant on ShopApiPlugin

NenadStef:
@lchrusciel @TiMESPLiNTER 

Hi Lukasz

As we talked some time ago I implemented two new constraints on ShopApiPlugin.

First constraint is for checking wether Product is enabled or disabled and the second one is doing the same for product variant. They are triggered on adding item in the cart.
I also added two spec tests for this two new validators.


lchrusciel (commented):
Hey Nenad,

thanks for your contribution. It looks good! However, can you add also functional tests with PHPUnit? I think you should add two tests, which will cover adding disabled product and disabled variant to the cart. Sample tests can be found here: https://github.com/Sylius/ShopApiPlugin/blob/master/tests/Controller/Cart/PutItemToCartApiTest.php

In another PR we could filter disabled variants from responses.

Also, my congratulations for your first PR 🎉

src/Validator/Constraints/ProductEligibility.php
```diff
@@ -0,0 +1,25 @@
...
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+
+final class ProductEligibility extends Constraint
+{
+    /** @var string */
+    public $message = 'sylius.shop_api.product.eligibility';
```
diimpp, line 12: How about  `sylius.shop_api.product.non_eligible` or `ineligible`

src/Validator/Constraints/ProductVariantEligibility.php
```diff
@@ -0,0 +1,25 @@
...
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+
+final class ProductVariantEligibility extends Constraint
+{
+    /** @var string */
+    public $message = 'sylius.shop_api.product_variant.eligibility';
```
diimpp, line 12: How about  `sylius.shop_api.product_variant.non_eligible` or `ineligible`

NenadStef:
@lchrusciel Thanks, I will add tests and for that second PR on which response do you think?

src/Validator/Product/ProductEligibilityValidator.php
```diff
@@ -0,0 +1,36 @@
...
+    /** {@inheritdoc} */
+    public function validate($productCode, Constraint $constraint): void
+    {
+        if (null === $productCode) {
+            return;
+        }
+
+        $product = $this->productRepository->findOneByCode($productCode);
+
+        if (null === $product || $product->isEnabled()) {
```
diimpp, line 30: Personally I would have separated those cases as one is "ignore" case and another is eligibility check, so it doesn't feel right to mix them

diimpp:
Another case --  when product was added to cart as eligible, but become ineligible over time and customer went to checkout/complete step.

Those cases are the same as out of stock and require same implementation. Might as well check for them too

src/Resources/config/validation/cart/PutOptionBasedConfigutableItemToCartRequest.xml
```diff
@@ -42,6 +42,7 @@
             </constraint>
```
diimpp, line 42: There is another request, which can benefit from this check

> src/Handler/Cart/ChangeItemQuantityHandler.php

NenadStef:
@lchrusciel @diimpp I have a bit of situation with running integration tests.

Because in composer.json is set that `sylius/sylius` must be ˆ1.4 running composer install will install sylius version 1.8 which will cause a lot of errors because of `SonataCoreBundle` which don't exists in 1.8 version. This can be fix by modifying `test/Application/config/bundles.php` by adding new bundles but in that case this pull request should be strictly for versions 1.8 and higher.

If I run integration tests with sylius version 1.4 I don't have bundles problem but this pull request has a problem because in 1.4 version there is no method `isEnabled` on `ProductVariant`.

What would be your suggestion?

mamazu:
It indeed looks like we don't have any tests in the matrix for the most recent version of ShopApi only up to 1.6. We probably should update it from the Plugin Skeleton. But this is a different issue.

Hm, if it is not defined in 1.4 we might need to raise the minimum required version of the `sylius/sylius` package.

NenadStef:
@lchrusciel @diimpp @mamazu 

This pull request is set strictly to run with `sylius/sylius ˆ1.8` so I did some modification to `tests/Application/config` files and `bundles.php`. All unit and integration tests are run successfully for this pull request.

As for product and product variant `enabled/disabled` functionality next cases are covered:

1. On adding item to the cart validation will be triggered to check if product or product variant is enabled or disabled
2. On changing quantity for order item validation will be triggered to check if product or product variant is enabled or disabled
3. On cart checkout validation will be triggered to check if product or product variant is enabled or disabled

This validators are covered with unit and integration tests.

In each of this new validators there must be a check to see if object for which we are running query is found otherwise exception would be thrown.
This validators do not handle missing objects because that is a  job for other validators.

Take a look at this pull request when you got time and tell me what you think about it.

mamazu:
We are currently planning a release for ShopAPI with support for 1.8 and onwards (#688). If you can wait with this pull request then you should be able to just rebase the branch onto the new master and it should work.

src/Validator/Cart/CartCheckoutItemEligibilityValidator.php
```diff
@@ -0,0 +1,62 @@
...
+
+            /** @var ProductVariantInterface $variant */
+            $variant = $item->getVariant();
+
+            if ($variant->isEnabled()) {
+
+                /** @var ProductInterface $product */
+                $product = $variant->getProduct();
+
+                if (!$product->isEnabled()) {
```
diimpp, line 48: `if` nesting should be avoided as good practice.

You can write them like so:
```php
    if (!$variant->isEnabled()) {
        $this->context->addViolation($constraint->messageOnNonEligibleCartItemVariant);
        
        break;
    }
    if (!$product->isEnabled()) {
        $this->context->addViolation($constraint->messageOnNonEligibleCartItem); 

        break;
    }
```

Additionally, when you validation whole cart, you need to specify which cart item is invalid

src/Validator/Cart/CartItemEligibilityValidator.php
```diff
@@ -0,0 +1,57 @@
...
+
+    public function __construct(OrderItemRepositoryInterface $orderItemRepository)
+    {
+        $this->orderItemRepository = $orderItemRepository;
+    }
+
+    /** {@inheritdoc} */
+    public function validate($id, Constraint $constraint): void
+    {
+        if (null === $id) {
```
diimpp, line 28: I think it's safe to leave this to `find($id)` to find nothing.

src/Validator/Cart/CartItemEligibilityValidator.php
```diff
@@ -0,0 +1,57 @@
...
+            return;
+        }
+
+        /** @var OrderItem|null $orderItem */
+        $orderItem = $this->orderItemRepository->find($id);
+
+        if ($orderItem) {
```
diimpp, line 35: Same as in previous validator. Please remove unnecessary empty lines and `if` nesting.

src/Validator/Constraints/CartItemEligibility.php
```diff
@@ -0,0 +1,28 @@
...
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+
+final class CartItemEligibility extends Constraint
+{
+    /** @var string */
+    public $messageOnNonEligible = 'sylius.shop_api.cart_item.product.non_eligible';
```
diimpp, line 12: If you have `$messageOnNonEligibleVariant`, then `$messageOnNonEligible` should be `$messageOnNonEligibleProduct`

Personally, I would go for `nonEligibleProductMessage` and `nonEligibleProductVariantMessage` instead

src/Validator/Constraints/CartCheckoutItemEligibility.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+
+final class CartCheckoutItemEligibility extends Constraint
```
diimpp, line 9: Same about name

diimpp (changes requested):

.gitignore
```diff
@@ -5,3 +5,6 @@
 
 .phpunit.result.cache
 /phpunit.xml
+
+/.idea/
+/docker/
```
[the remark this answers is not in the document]
mamazu: This is environment specific and should not be contained in the .gitignore. You can set up a global .gitignore to not worry about it anymore.

More info here: https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

src/Validator/Cart/CartCheckoutItemEligibilityValidator.php
NenadStef: @diimpp In this request only value that is pass is token, so do you have some example where you build path to properties that are not present in the request?

NenadStef:
@mamazu When do you expect to release new version of the ShopApiPlugin that will be compatible with sylius/sylius 1.8 so we can plan whether to wait for the release and this pull request or just to use our fork.

mamazu:
As far as I know @lchrusciel is working on it currently with an estimated release date to be next week-ish.

NenadStef:
@TiMESPLiNTER @lchrusciel @mamazu @diimpp 

Regarding previous work that is done on this branch and also comments on https://github.com/Sylius/ShopApiPlugin/issues/683 this is a quick summary what is done so far:

New constraints 

1. On `AddCouponRequest`
	- CartEmpty constraint

2. On `ChangeItemQuantityRequest`
	- CartEmpty constraint
	- CartItemEligibility constraint

3. On `EstimateShippingCoastRequest`
	- CartEmpty constraint

4. On `PutOptionBasedConfigutableItemToCartRequest`
	- ProductEligibility constraint

5. On `PutSimpleItemToCartRequest`
	- ProductEligibility constrain

6. On `PutVariantBasedConfigutableItemToCartRequest`
	- ProductEligibility constrain
	- ProductVariantEligibility constrain

7. On `AddressOrderRequest`
	- CartEmpty constraint

8. On `ChoosePaymentMethodRequest`
	- CartEmpty constraint

9. On `CompleteOrderRequest`
	- CartEmpty constraint	
	- CartEligibility constraint

All this validators and cases are covered with unit and integration tests.

**TODO**:
	Filter disabled products and product variants from responses on
1. `sylius_shop_api_product_show_latest`
2. `sylius_shop_api_product_show_catalog_by_slug`
3. `sylius_shop_api_product_show_catalog_by_code`

Once this branch is updated with updated version that is compatible with sylius 1.8 you will be able to do a proper code review.

lchrusciel:
v1.2 was just released, I will try to review this PR in the upcoming days.

mamazu:
The new version is out. Please rebase your branch to the new master.

About your change log, I would say that adding a `CartNotEmpty` shouldn't be part of this pull request and rather a separate one. (It shouldn't be much to change anyways I suppose)

src/Validator/Cart/CartCheckoutItemEligibilityValidator.php
diimpp: @NenadStef sorry forgot to answer -- you can specify any path in the violation, validation doesn't care if path exists or not and front-end will be able to parse if path will be similar to cart/summary output.

So it's more of conceptual question rather than technical. Like "Is it alright to specify path from summary endpoint".
Personally I would go for it, because usually front-end needs to indicate which item causes issue

NenadStef:
@lchrusciel Travis is failing because this pull request should be for sylius 1.8 and higher (it fails for sylius 1.7). Product variant isEnabled method was added in sylius 1.8.

NenadStef:
@lchrusciel @TiMESPLiNTER @mamazu @diimpp 

Regarding filtration of the disabled product or product variants I need some clarification.

1. If product is disabled it should not be in shown product list
2. If is a simple product (1 product and 1 variant) and product is enabled but product variant is disabled product should not be shown in product list
3. If product has more then one variant and product is enabled but one of the variant is disabled product should be shown with only enabled variants

Are this assumptions correct?

**Additional validation**

Is there a need to add additional product and product variant eligibility validation  on request such as add product review or show product details...

mamazu:
Sounds good to me.

lchrusciel:
Sorry, for the late feedback loop, I will try to jump into this PR next week. 

According to supported versions, the last release supported 1.7 & 1.8. As 1.7 has security fixes only, I would vote for bumping minimal requirement to 1.8.

NenadStef:
@lchrusciel 

Hi Lukasz

I added filtration for disabled product and product variants on routes for latest products, product catalogs and product details.

Disabled products are handled automatically in Sylius database queries but for disabled product variants I added filtration so now only enabled product variants will be shown.

Regarding disabled product variant I have a question how this works in Sylius core. In case of simple product when product is set to disabled related product variant is also set to disabled, but when product variant is set to disabled related product is not set as disabled. In case when product has multiple variants we can set all product variants to disabled and product will be left as enabled.
From the shop api plugin this will be manifest in form that enabled products will be shown only with enabled variants and if all variants are disabled it will be shown as product without variants.

I don’t know if this behaviour for disabled products and product variants is what was intended but if not it can be change with another pull request on Sylius core.

src/Validator/Cart/CartEligibilityValidator.php
```diff
@@ -66,9 +68,9 @@ public function validate($token, Constraint $constraint): void
 
             Assert::isInstanceOf($product, ProductInterface::class);
             if (!$product->isEnabled()) {
-                $this->context->addViolation(
-                    $constraint->nonEligibleCartItemMessage
-                );
+                $this->context->buildViolation($constraint->nonEligibleCartItemMessage)
+                    ->atPath('items[' . $key . '].product.code')
+                    ->addViolation();
```
NenadStef, line 74: @diimpp For the cart eligibility I mapped properties for the product and product variant in this way.

lchrusciel (commented):
Hey Nenad, due to Christmas time it took a little bit more time to go through your PR. 

According to your assumptions, that you've written above you were almost correct. From the three points, only one that you've asked again works slightly differently. I mean, the case, where you have enabled product together with the disabled variant. In pure Sylius, you may have a product without a variant. This will be handled in the shop with the product that is out of stock and you are not able to buy it. The same should happen in our case. 

The only important thing I would like to clarify is, this 500 for product variant not found. Can we trigger 400 there? Also, the build is red

Overall, thanks for the great work. Maybe we could think about tagging a new release just after polishing these features?

tests/Responses/Expected/cart/product_variant_has_not_been_found_response.json
```diff
@@ -1,4 +1,4 @@
 {
-    "code": 404,
-    "message": "Variant not found for given configuration"
+    "code": 500,
+    "message": "Variant could not be resolved"
```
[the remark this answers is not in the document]
NenadStef: Hi Lukasz, happy holidays.

If I understand correctly status code 500 is return because product variant is not found. If you look at `tests/Controller/Cart/PutItemToCartApiTest.php` in `it_throws_an_exception_if_product_variant_has_not_been_found` method for this test is using option based product and request in this test is not formatted correctly on purpose (It has only one of two options) so in this case it should raise 500 error

src/Validator/Constraints/CartNotEmpty.php
```diff
@@ -0,0 +1,25 @@
...
+declare(strict_types=1);
+
+namespace Sylius\ShopApiPlugin\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+
+final class CartNotEmpty extends Constraint
+{
+    /** @var string */
+    public $emptyCartMessage = 'sylius.shop_api.checkout.cart.empty';
```
lchrusciel, line 12: Can we be a little bit more coherent?
```suggestion
    public $message = 'sylius.shop_api.checkout.cart.empty';
```

tests/Responses/Expected/cart/product_variant_has_not_been_found_response.json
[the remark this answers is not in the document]
lchrusciel: For me, the app shall behave exactly the same way in both cases. This request seems to be just bad, wrong data. It should not result in sever error IMO.

NenadStef:
@lchrusciel Hi Lukasz

So I wanted to inform you about progress on this pull request.

1. That 500 error is solved by adding additional validator for checking if option base product variant exists.
2. Also additional validator is added for checking if option base product variant is eligible
3. All code is refactor so now it can work both with `Sylius 1.7` and `Sylius 1.8`  
4. All spec test are refactored so now thy are working both on 1.7 and 1.8
5. All integration tests are working on 1.7 and 1.8
6. Static analysis is without errors

**Help needed**

One last thing that is still failing on Travis (see [stats](https://github.com/Sylius/ShopApiPlugin/pull/684/checks?check_run_id=1888740799) and [build](https://travis-ci.com/github/Sylius/ShopApiPlugin/jobs/482659165)) is coding standard. I could really use your help to solving this.

lchrusciel:
Hey Nenad, I'm hard for me to say this, but I've never asked you to support both versions at the same time... 

Your build was blocked for a really long time and I'm sorry for that. Today, I finally dropped support for Sylius 1.7 and this PR could go on. Nonetheless, all integration code has to be removed. Also, the build should be green now, as we moved to GitHub actions. In terms of ECS problems, it was removed from the CI pipeline.

NenadStef:
@lchrusciel I saw that you did update for Travis and I merge it on this pull request. Regarding versions compatibility I did that on my own because there was no answer about version compatibility for a long time but I can change it back so it can only be run on version 1.8.

lchrusciel:
Once again, I'm sorry for being a bottleneck in this manner. I truly appreciate your work and would like to merge it to the core, once we will remove the 1.7 compatibility layer

NenadStef:
@lchrusciel It's no problem I will remove that compatibility stuff in next couple of days

NenadStef:
@lchrusciel I removed all Sylius 1.7 compatibility files and now pull-request is only for 1.8 version, so take a look and tell me what you think.

NenadStef:
@lchrusciel Hi Lukasz

I was wondering is there any problem with Travis? It's been pending for couple of days.

mamazu:
We moved from Travis to github actions. If you rebase your branch to the current master you will get that too.

src/Factory/Product/ListProductViewFactory.php
```diff
@@ -54,7 +54,9 @@ private function createWithVariants(ProductInterface $product, ChannelInterface
         /** @var ProductVariantInterface $variant */
         foreach ($product->getVariants() as $variant) {
             try {
-                $productView->variants[$variant->getCode()] = $this->variantViewFactory->create($variant, $channel, $locale);
+                if ($variant->isEnabled()) {
```
mamazu, line 57: The `if` statement can be outside of the `try`-Block.

mamazu: Maybe even filter the product variants before iterating over it? Because then you wouldn't need if statements here and in line 70.

mamazu (commented):
This is a very huge merge request and some of the features could be added as a stand alone merge request. For example the "CartNotEmpty" validator, tests and configuration.

NenadStef:
@mamazu Hi mamazu, thanks for all the feedback.

I updated my fork with your's master branch but still having Travis CI pending as required check.

I know that this is a huge PR but it can be summarize to this

1. Cart not empty validation
2. Option base product exists validation
3. Product eligibility (simple, variant, option base) validation and build view

I understand need to split this PR in small chunks, it would be easy to review it but thing is that this PR is mostly intended to be used by the company that I work for and I was hoping that it will be approved soon.

If you insist to go with separating PR to small chunks I will have to talk with my team at company to see how much time I will have to work on this because we are having very intense sprints in next couple of weeks.

@lchrusciel What do you think about splitting this PR in small chunks?

NenadStef:
@lchrusciel Hi Lukasz

Do you have any feedback about how are we going to handle this pull request?

lchrusciel (approved):

lchrusciel:
Hey Nenad,

thanks a lot for keeping this PR alive and sorry for my late feedback. In addition, thanks Max and Dmitri for your early review.

NenadStef:
@lchrusciel Hey Lukasz, thanks for approving this

PR #687, opened 2020-11-11
Fixing cart item definition

### LINKED ISSUE #641: Incorrect product variants documentation
Hey everyone,

The documentation states that when retrieving the a cart /carts/{token} should return the variants with the variant code as the key of the array. This is not happening and the key is the index of the array.

mamazu:
Fixes #641

### DIFF
--- a/doc/swagger.yml
+++ b/doc/swagger.yml
@@ -1326,25 +1326,21 @@ definitions:
                 type: "integer"
                 example: 1000
             product:
-                $ref: "#/definitions/Product"
+                type: "object"
+                allOf:
+                  - $ref: "#/definitions/ProductMinimal"
+                properties:
+                  variants:
+                    type: "array"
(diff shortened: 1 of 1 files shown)


doc/swagger.yml
```diff
@@ -1326,47 +1326,71 @@ definitions:
...
+                type: "string"
+                example: "A mug."
+            metaKeywords:
+                type: "string"
+                example: "mug drink logan"
+            metaDescription:
+                type: "string"
+                example: "Some meta description"
             images:
+                type: "string"
```
lchrusciel, line 1387: isn't it array?

lchrusciel (approved):

doc/swagger.yml
mamazu: Ah yes, thanks for catching it.

PR #700, opened 2021-02-21
Adding a subtotal property to the cart endpoint

mamazu:
This partially solves #638 but without the BC break.


lchrusciel (approved):

lchrusciel:
Thanks, @mamazu! :tada:

PR #648, opened 2020-05-20
Add availibility check for product

alexander-schranz:
fixes #594

TODO:

 - [ ] Tests

PR #701, opened 2021-02-21
Availability product variant

mamazu:
Replaces #648

### DIFF
--- a/src/Modifier/OrderModifier.php
+++ b/src/Modifier/OrderModifier.php
@@ -36,7 +36,7 @@ final class OrderModifier implements OrderModifierInterface
         OrderItemQuantityModifierInterface $orderItemQuantityModifier,
         OrderProcessorInterface $orderProcessor,
         ObjectManager $orderManager,
-        ?AvailabilityCheckerInterface $availabilityChecker
+        ?AvailabilityCheckerInterface $availabilityChecker = null
     ) {
         $this->cartItemFactory = $cartItemFactory;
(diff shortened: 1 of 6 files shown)


lchrusciel (approved):
Seems ok, but probably validation to expose nice 400 error to the customer should be added as well

lchrusciel:
Thanks, @mamazu! :tada:

mamazu:
The idea behind this was that it was optional to add this feature. I wasn't sure it it would be a BC break to do so.

src/Modifier/OrderModifier.php
```diff
@@ -70,4 +85,15 @@ private function getCartItemToModify(OrderInterface $cart, ProductVariantInterfa
 
         return null;
     }
+
+    private function checkCartQuantity(ProductVariantInterface $productVariant, int $targetQuantity): void
+    {
+        if($this->availabilityChecker === null) {
```
diimpp, line 91: Missing whitespace between `if` and `(`. ecs should have caught that, strange.

diimpp:
Missing DI definition with actual service injection.
I think this is not a BC, it's a part of standard sylius behavior and only expected to function this way.

We should check for enabled/disabled over here as well.

Maybe we can introduce a point of extension over here, ShopApi AvailabiltyCheckerInterface with CompositeAvailabilityChecker, which will check for Inventory availability, enabled/disabled and common use case for my projects -- archived products/variants

Sounds good? I already have such system in place for my project, I can do follow PR.

PR #711, opened 2021-04-22
Adding an assertion that the payment method needs to be available.

### LINKED ISSUE #650: [BUG] Missing validation during choosing cart payment method
Example: Lets say we have Method_a and Method_b in the Sylius. But only Method_a is allowed for the channel.

We call [get] /checkout/XXXXXXX/payment and receive Method_a as the only available option. What is correct.

Then we [put] /checkout/XXXXXXX/payment/0 with the body: {"method": "Method_b"}  ( .._b not .._a)

And it works. Method_b becomes payment method for our order what should not happen.

So there should be some validation for this situation.

mamazu:
Fixes #650

### DIFF
--- a/src/Handler/Cart/ChoosePaymentMethodHandler.php
+++ b/src/Handler/Cart/ChoosePaymentMethodHandler.php
@@ -6,10 +6,12 @@ namespace Sylius\ShopApiPlugin\Handler\Cart;
 
 use SM\Factory\FactoryInterface;
 use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Core\Model\PaymentInterface;
 use Sylius\Component\Core\Model\PaymentMethodInterface;
 use Sylius\Component\Core\OrderCheckoutTransitions;
 use Sylius\Component\Core\Repository\OrderRepositoryInterface;
 use Sylius\Component\Core\Repository\PaymentMethodRepositoryInterface;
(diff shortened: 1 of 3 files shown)


src/Handler/Cart/ChoosePaymentMethodHandler.php
```diff
@@ -24,36 +26,56 @@ final class ChoosePaymentMethodHandler
...
 
+    /** @var PaymentMethodsResolverInterface */
+    private $paymentMethodsResolver;
+
     public function __construct(
         OrderRepositoryInterface $orderRepository,
         PaymentMethodRepositoryInterface $paymentMethodRepository,
-        FactoryInterface $stateMachineFactory
+        FactoryInterface $stateMachineFactory,
+        PaymentMethodsResolverInterface $paymentMethodsResolver
```
lchrusciel, line 36: This argument should be nullable and should trigger deprecation error if not send

mamazu: Done. If the property is not set then the validation (aka assertion) will not be triggered as well. So fully backwards compatible.

mamazu:
Then this could be merged.

lchrusciel:
Thank you, @mamazu! :tada:

lchrusciel:
Thanks Max 🏅

PR #715, opened 2021-06-21
Fix Symfony package constraints

TiMESPLiNTER:
With the current composer.json constraints regarding the Symfony packages we only are allowed to install version `5.2.0` of the packages. This PR fixes this and allows `^5.2` to be installed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
         "php": "^7.3",
         "sylius/sylius": "^1.8",
         "lexik/jwt-authentication-bundle": "^2.5",
-        "symfony/messenger": "^4.4|5.2",
+        "symfony/messenger": "^4.4|^5.2",
         "doctrine/doctrine-bundle": "^2.0",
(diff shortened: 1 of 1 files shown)


vvasiloi (approved):

mamazu:
Could you try to fix the pipeline. It seems like we are missing some typehints.

TiMESPLiNTER:
> Could you try to fix the pipeline. It seems like we are missing some typehints.

Not so easy imo. For example this is a problem:

```php
        foreach ($request->request->get('items') as $item) {
            $item['token'] = $token;
            $commandRequests[] = $this->provideCommandRequest($item);
        }
```

it's not about type hints but to properly fix it imo we would need to do something like this:

```php
        $items = $request->request->get('items');

        if (false === is_iterable($items)) {
            throw new BadRequestHttpException('items need to be an array');
        }

        foreach ($items as $item) {
            $item['token'] = $token;
            $commandRequests[] = $this->provideCommandRequest($item);
        }
```

mamazu:
Yeah true that. Theoretically the api would currently also accept a string in that position.

PR #717, opened 2021-06-29
[Maintenance] Support for Sylius 1.8 dropped, Sylius 1.10 added,  Symfony 5.3 added, PHP 8.0 added

lchrusciel:
Replacing #715

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
                 php: [7.4, 7.3]
                 node: [10.x]
                 mysql: [5.7, 8.0]
-                symfony: [^4.4, ^5.2]
+                symfony: [^4.4, ^5.2, ^5.3]
                 sylius: [~1.8.0, ~1.9.0]
 
                 exclude:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,19 +18,15 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: [7.4, 7.3]
+                php: [7.4, 8.0]
                 node: [10.x]
                 mysql: [5.7, 8.0]
-                symfony: [^4.4, ^5.2, ^5.3]
-                sylius: [~1.8.0, ~1.9.0]
-
+                symfony: [^4.4, ~5.2.0, ~5.3.0]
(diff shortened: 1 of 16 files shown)


.github/workflows/build.yml
```diff
@@ -18,19 +18,11 @@ jobs:
...
-                sylius: [~1.8.0, ~1.9.0]
-
-                exclude:
-                    - # PHP 7.3 does not support "caching_sha2_password" authentication plugin which is a default one in MySQL 8.0
-                        php: 7.3
-                        mysql: 8.0
-                    -
-                        sylius: ~1.8.0
-                        symfony: ^5.2
+                symfony: [^4.4, ^5.2, ^5.3]
```
GSadee, line 24: I guess that there will be the same version of Symfony installed for `^5.2` and `^5.3`
```suggestion
                symfony: [^4.4, ~5.2.0, ~5.3.0]
```

.github/workflows/build.yml
```diff
@@ -18,19 +18,11 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: [7.4, 7.3]
+                php: [7.4]
```
GSadee, line 21: What about PHP 8? Did you consciously omit it in this PR?

composer.json
```diff
@@ -5,9 +5,9 @@
     "license": "MIT",
     "require": {
         "php": "^7.3",
```
GSadee, line 7: If you would like to support PHP 8, it should be added also here:
```suggestion
        "php": "^7.3|^8.0",
```

.github/workflows/build.yml
```diff
@@ -18,19 +18,19 @@ jobs:
...
+                symfony: [^4.4, ~5.2.0, ~5.3.0]
+                sylius: [~1.9.0, ~1.10.0]
+
+            exclude:
+                - # Sylius 1.9 does not support PHP 8.0
+                    sylius: ~1.9.0
+                    php: 8.0
+                -
+                    sylius: ~1.8.0
+                    symfony: ^5.2
```
GSadee, line 33: As Sylius 1.8 is not supported:
```suggestion
```

GSadee (approved):

GSadee:
Thank you, Łukasz! :tada:

PR #723, opened 2022-02-05
Adding the license block to the coding standard

mamazu:
All php files should have a license block, so the easiest solution is to add a rule in the `ecs` for it.


Roshyo (approved):

lchrusciel:

PR #729, opened 2022-06-29
Updating the codestyle

mamazu:
## Why
We are no longer supporting Sylius 1.9 and this means that we can now require the codestyle plugin in a more recent version. This makes the codestyle stick more closely to the guides that the main repository provides.

## What
* I raised the default codestyle version to from ^3.0 to ^4.0 which is also the minimum version in 1.10.
* This then also allows us to migrate the codestyle settings to PHP (or in this case just take the settings from Sylius/Sylius and change the paths)


composer.json
```diff
@@ -51,5 +51,10 @@
         "branch-alias": {
             "dev-master": "1.0-dev"
         }
+    },
+    "config": {
+        "allow-plugins": {
+            "dealerdirect/phpcodesniffer-composer-installer": true
```
lchrusciel, line 57: Is it really needed for ECS to work?

lchrusciel (approved):

composer.json
mamazu: Nope, has been removed. But composer is so annoying without those things.

composer.json
```diff
@@ -32,11 +32,11 @@
     "scripts": {
         "analyse": [
             "@composer validate --strict",
-            "vendor/bin/ecs check --ansi --no-progress-bar spec src tests/Controller tests/DataFixtures tests/Request",
+            "vendor/bin/ecs check --ansi --no-progress-bar",
             "vendor/bin/phpstan analyse -c phpstan.neon --no-progress src"
         ],
         "fix": [
-            "vendor/bin/ecs check --ansi --no-progress-bar spec src tests/Controller tests/DataFixtures tests/Request --fix"
+            "vendor/bin/ecs check --ansi --no-progress-bar"
```
lchrusciel:
```suggestion
            "vendor/bin/ecs check --ansi --no-progress-bar --fix"
```

lchrusciel (approved):

lchrusciel (approved):

PR #734, opened 2022-12-08
Add Support for Sylius 1.12

TheMilek:
Changes:

- Dropped `symfony/web-server-bundle` as is no longer supported by Symfony
- Bumped up the `phpstan/phpstan` to 1.8 version and `phpstan/phpstan-webmozart-assert` to 1.2
- Updated `.env.dist` files to make them usable with the available documentation

**Important!**
To potential merge after merging https://github.com/Sylius/ShopApiPlugin/pull/735 and https://github.com/Sylius/ShopApiPlugin/pull/737


spec/Http/RequestBasedLocaleProviderSpec.php
```diff
@@ -47,20 +47,6 @@ function it_provides_a_locale_code(
         $this->getLocaleCode($request)->shouldReturn('fr_FR');
     }
 
-    function it_provides_a_locale_code_even_if_it_is_not_defined_explicitly(
```
TheMilek, line 50: IMHO this test is broken and redundant as `getLocale()` method cannot return null which also causes the error to occur.
[REF](https://github.com/symfony/http-foundation/blame/6.2/Request.php#L1368-L1371)

.github/workflows/build.yml
```diff
@@ -18,17 +18,18 @@ jobs:
...
-                mysql: [5.7, 8.0]
-                symfony: [^4.4, ~5.2.0, ~5.3.0]
-                sylius: [~1.10.0, ~1.11.0]
-
+                php: ["8.0", "8.1"]
+                symfony: ["~5.4.0"]
+                sylius: ["~1.11.10", "~1.12.0"]
+                node: ["16.x"]
+                mysql: ["5.7", "8.0"]
+
```
coldic3:
```suggestion

```

.github/workflows/build.yml
```diff
@@ -63,36 +64,23 @@ jobs:
             -
                 name: Output PHP version for Symfony CLI
                 run: php -v | head -n 1 | awk '{ print $2 }' > .php-version
-
-            -
+ 
+            -
```
coldic3:
```suggestion

            -
```

.github/workflows/build.yml
```diff
@@ -63,36 +64,23 @@ jobs:
...
+            -   
                 name: Restrict Symfony version
                 if: matrix.symfony != ''
                 run: |
                     composer global config --no-plugins allow-plugins.symfony/flex true
-                    composer require --no-progress --no-scripts --no-plugins "symfony/flex:^1.10"
+                    composer global require --no-progress --no-scripts --no-plugins "symfony/flex:^1.10"
                     composer config extra.symfony.require "${{ matrix.symfony }}"
-
+
```
coldic3:
```suggestion

```

composer.json
```diff
@@ -5,24 +5,24 @@
     "license": "MIT",
     "require": {
         "php": "^8.0",
-        "sylius/sylius": "^1.10",
+        "sylius/sylius": "~1.11.10 || ~1.12.0",
         "lexik/jwt-authentication-bundle": "^2.5",
-        "symfony/messenger": "^4.4|^5.2",
```
coldic3: We usually drop SF 4.4 in a separate PR. I think it would be nice to do the same in this case as well.

phpstan.neon
```diff
@@ -6,16 +6,18 @@ parameters:
     checkGenericClassInNonGenericObjectType: false
     checkMissingIterableValueType: false
 
-    excludes_analyse:
-        # Makes PHPStan crash
+    excludePaths:
         - src/DependencyInjection/Configuration.php
 
+    treatPhpDocTypesAsCertain: false
```
coldic3: Why?

phpstan.neon
```diff
@@ -6,16 +6,18 @@ parameters:
...
-        - '/Call to static method Webmozart\\Assert\\Assert::implementsInterface\(\) with string/'
         - '/Empty array passed to foreach./'
         - '/Property Sylius\\ShopApiPlugin\\Request\\Checkout\\AddressOrderRequest::\$\w+Address \(array\|null\) does not accept bool\|float\|int\|string\|null\./'
         - '/Property Sylius\\ShopApiPlugin\\Request\\Customer\\UpdateCustomerRequest::\$birthday \(DateTimeImmutable\|null\) does not accept bool\|float\|int\|string\|null\./'
         - '/Parameter #3 \$options of class Sylius\\ShopApiPlugin\\Request\\Cart\\PutOptionBasedConfigurableItemToCartRequest constructor expects array\|null, bool\|float\|int\|string\|null given\./'
         - '/Argument of an invalid type bool\|float\|int\|string supplied for foreach, only iterables are supported\./'
+        - '/^Property Sylius\\ShopApiPlugin\\Request\\Cart\\PutSimpleItemToCartRequest\:\:\$productCode \(string\) on left side of \?\? is not nullable\./'
+        - '/^Property Sylius\\ShopApiPlugin\\Request\\Cart\\PutVariantBasedConfigurableItemToCartRequest\:\:\$productCode \(string\) on left side of \?\? is not nullable\./'
```
coldic3:
```suggestion
        - '/^Property Sylius\\ShopApiPlugin\\Request\\Cart\\(\w+)\:\:\$productCode \(string\) on left side of \?\? is not nullable\./'
```

src/Request/Cart/PutOptionBasedConfigurableItemToCartRequest.php
```diff
@@ -22,7 +22,7 @@ class PutOptionBasedConfigurableItemToCartRequest implements RequestInterface
     protected $token;
 
     /** @var string */
-    protected $productCode;
+    protected $productCode = '';
```
coldic3: Is it necessary?

src/Request/Customer/UpdateCustomerRequest.php
```diff
@@ -40,7 +40,7 @@ class UpdateCustomerRequest implements ShopUserBasedRequestInterface
     protected $phoneNumber;
 
     /** @var bool */
-    protected $subscribedToNewsletter;
+    protected $subscribedToNewsletter = false;
```
coldic3: Is it necessary?

tests/Controller/Checkout/CompleteOrderApiTest.php
```diff
@@ -36,12 +38,17 @@
...
+    use MailerAssertionsTrait;
 
     /**
      * @test
      */
     public function it_allows_to_complete_checkout(): void
     {
+        if ($this->isTestExecuatable()) {
+            $this->markTestSkipped('This test should be executed only with Symfony Mailer.');
+        }
```
coldic3: Is it needed?

tests/Controller/Customer/RegisterApiTest.php
```diff
@@ -55,9 +56,9 @@ public function it_allows_to_register_in_shop_and_sends_a_verification_email_if_
         Assert::assertNotNull($user);
         Assert::assertFalse($user->isEnabled());
 
-        /** @var EmailCheckerInterface $emailChecker */
-        $emailChecker = $this->get('sylius.behat.email_checker');
-        Assert::assertTrue($emailChecker->hasRecipient('vinny@fandf.com'));
+        self::setContainer($this->client->getContainer());
```
coldic3: You can do it in the `setUp` method.

tests/Controller/Utils/MailerAssertionTrait.php
```diff
@@ -0,0 +1,59 @@
...
+            return $chosenContainer->get('mailer.logger_message_listener')->getEvents();
+        }
+
+        throw new AssertionFailedError('A client must have Mailer enabled to make email assertions. Did you forget to require symfony/mailer?');
+    }
+
+    protected static function setContainer(ContainerInterface $container): void
+    {
+        self::$chosenContainer = $container;
+    }
```
coldic3: In traits, you can rely on what will be present in the object that uses the trait. To do so, you can use the `@mixin` annotation above trait in line 20, eg.:
```
@mixin \Tests\Sylius\ShopApiPlugin\Controller\JsonApiTestCase
```
to have everything hinted by IDE. Then instead of using `setContainer` you can just use `$this->client->getContainer()`.

composer.json
TheMilek: I agree t is done in separate PR, but SF 4.4 isn't working here, so I removed it also here to green the tests as fixing it is pointless IMO [REF](https://github.com/Sylius/ShopApiPlugin/pull/735)

tests/Controller/Checkout/CompleteOrderApiTest.php
TheMilek: sadly yes, in other case I have to make a separate build only for swiftmailer, as the cache directory is different

phpstan.neon
TheMilek: I agree this one is redundant as to fix the eror it just needed to change order of the statements

src/Request/Cart/PutOptionBasedConfigurableItemToCartRequest.php
TheMilek: I wanted to fix the "left side of ?? is not nullable" problem by using that, but it seems to be made on purpose like that to use it in validation. This change I made here brokes it.

src/Request/Customer/UpdateCustomerRequest.php
TheMilek: as above

tests/Controller/Checkout/CompleteOrderApiTest.php
TheMilek: after call with @coldic3 this statement had been reworked

mamazu (approved):
Thanks for the work you put in to get this plugin up to speed.

I just have one two things:
Would you mind summarizing in a comment what requirements have changed. I see there is no longer support for Symfony 4, we also dropped Sylius 1.10 anything I missed? 
Furthermore, do we need a change log entry for the upgrade or should it continue to work out of the box?

TheMilek:
> Thanks for the work you put in to get this plugin up to speed.
> 
> I just have one two things: Would you mind summarizing in a comment what requirements have changed. I see there is no longer support for Symfony 4, we also dropped Sylius 1.10 anything I missed? Furthermore, do we need a change log entry for the upgrade or should it continue to work out of the box?

Hi @mamazu!
Yeah it should work out of the box, also with these updated .env files the steps in the `Readme` file are the same. I updated the description with the changes as You requested 💃🏽. The continuation of this PR is [here](https://github.com/Sylius/ShopApiPlugin/pull/736) where I also add SF6 support

jakubtobiasz (approved):

mamazu:
@TheMilek That's great. Then I'll merge it. Never mind. Hubkit isn't working. Can you merge it?

TheMilek:
@mamazu I just let know my teammates from Sylius 😄

tests/Application/.env.dist
```diff
@@ -21,3 +21,8 @@ DATABASE_URL=mysql://root@127.0.0.1/sylius_%kernel.environment%?serverVersion=5.
 # Delivery is disabled by default via "null://localhost"
 MAILER_URL=smtp://localhost
 ###< symfony/swiftmailer-bundle ###
+MESSENGER_TRANSPORT_DSN=sync://
```
coldic3, line 26:
```suggestion
###> symfony/messenger ###
MESSENGER_TRANSPORT_DSN=sync://
###< symfony/messenger ###
```

It doesn't matter a lot but if we used to mark which env var is from which Symfony component it can also be done here to be consistent.

tests/Controller/Checkout/CompleteOrderApiTest.php
```diff
@@ -86,6 +92,62 @@ public function it_allows_to_complete_checkout(): void
         $response = $this->complete($token, $data);
         $this->assertResponseCode($response, Response::HTTP_NO_CONTENT);
 
+        $this->assertEmailCount(1);
```
coldic3: I see this is the only difference between `it_allows_to_complete_checkout_with_swiftmailer` and `it_allows_to_complete_checkout`. Therefore maybe the better idea is to have one test with some custom assertion that checks if SF mailer or swiftmailer is used and uses the right assertion. WDYT?

tests/Controller/Checkout/CompleteOrderApiTest.php
```diff
@@ -629,4 +691,25 @@ private function complete(string $token, ?string $data = null): Response
...
+    }
+
+    private function isSwiftMailerAvailable(): bool
+    {
+        if (self::$clientContainer->has('swiftmailer')) {
+            return self::$clientContainer->has('swiftmailer');
+        }
+
+        return false;
+    }
```
coldic3: Regarding the comment above it could be:

```php
    private function assertEmailCount(int $count): bool
    {
        if ($this->isSymfonyMailerAvailable()) {
            return parent::assertEmailCount($count);
        }

        $mailCollector = $this->client->getProfile()->getCollector('swiftmailer');
        return $this->assertSame($count, $mailCollector->getMessageCount());
    }
```

Not sure it's a working code OFC but it could be something similar :grinning:

coldic3: And you can move it to the `Tests\Sylius\ShopApiPlugin\Controller\Utils\MailerAssertionsTrait`

TheMilek: https://github.com/Sylius/ShopApiPlugin/pull/734/commits/0e914d6c0650776979aad87ef977629be082867d
what do you think? I think doing it in trait is the same effort but less time consuming as in trait I can't get client directly

.github/workflows/build.yml
```diff
@@ -18,17 +18,18 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: [8.0, 8.1]
-                node: [10.x]
-                mysql: [5.7, 8.0]
-                symfony: [^4.4, ~5.2.0, ~5.3.0]
-                sylius: [~1.10.0, ~1.11.0]
+                php: ["8.0", "8.1"]
+                symfony: ["~5.4.0"]
```
GSadee:
```suggestion
                symfony: ["^5.4"]
```

composer.json
```diff
@@ -5,24 +5,24 @@
...
+        "symfony/messenger": "^4.4 || ^5.4",
```
GSadee: You are dropping Symfony 4.4 in builds, but not in the composer, remove the support also here and it would be best to do it in the separate PR that you have prepared. 
The same with Sylius 1.10, it would be best to remove support for it in a separate PR, probably not now, but I leave a comment for the future.

composer.json
```diff
@@ -56,6 +56,9 @@
     "extra": {
         "branch-alias": {
             "dev-master": "1.0-dev"
```
GSadee:
```suggestion
            "dev-master": "1.6-dev"
```

.github/workflows/build.yml
```diff
@@ -18,17 +18,18 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: [8.0, 8.1]
-                node: [10.x]
-                mysql: [5.7, 8.0]
-                symfony: [^4.4, ~5.2.0, ~5.3.0]
-                sylius: [~1.10.0, ~1.11.0]
+                php: ["8.0", "8.1"]
+                symfony: ["^5.2"]
```
coldic3:
```suggestion
                symfony: ["^5.4"]
```
as Sylius 1.11 and 1.12 does not support 5.2 and 5.3 SF versions.

composer.json
```diff
@@ -5,24 +5,24 @@
     "license": "MIT",
     "require": {
         "php": "^8.0",
-        "sylius/sylius": "^1.10",
+        "sylius/sylius": "~1.11.10 || ~1.12.0",
         "lexik/jwt-authentication-bundle": "^2.5",
-        "symfony/messenger": "^4.4|^5.2",
+        "symfony/messenger": "^5.2",
```
coldic3:
```suggestion
        "symfony/messenger": "^5.4",
```
and the same for the rest of SF components

composer.json
```diff
@@ -55,7 +55,10 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.0-dev"
+            "dev-master": "1.6-dev"
+        },
+        "symfony": {
+            "require": "^5.2"
```
coldic3:
```suggestion
            "require": "^5.4"
```

tests/Controller/Utils/MailerAssertionsTrait.php
```diff
@@ -0,0 +1,96 @@
...
+        if (self::isSymfonyMailerAvailable()) {
+            self::assertThat(self::getMessageMailerEvents(), new Constraint\EmailCount($count, $transport), $message);
+        }
+
+        if (self::isSwiftMailerAvailable()) {
+            $mailCollector = $this->client->getProfile()->getCollector('swiftmailer');
+            self::assertSame($count, $mailCollector->getMessageCount());
+        }
+
+        if (!self::isSymfonyMailerAvailable() && !self::isSwiftMailerAvailable()) {
```
coldic3: You can add an early return to the if statements above and then remove the 3rd if statement.

coldic3: And I would be for throwing an Exception here.

composer.json
TheMilek: https://github.com/Sylius/ShopApiPlugin/pull/738

composer.json
TheMilek: https://github.com/Sylius/ShopApiPlugin/pull/738

.github/workflows/build.yml
TheMilek: https://github.com/Sylius/ShopApiPlugin/pull/738

tests/Controller/Customer/RegisterApiTest.php
```diff
@@ -55,16 +60,19 @@ public function it_allows_to_register_in_shop_and_sends_a_verification_email_if_
         Assert::assertNotNull($user);
         Assert::assertFalse($user->isEnabled());
 
-        /** @var EmailCheckerInterface $emailChecker */
-        $emailChecker = $this->get('sylius.behat.email_checker');
-        Assert::assertTrue($emailChecker->hasRecipient('vinny@fandf.com'));
+        $email = self::getMailerMessage();
+        self::assertEmailAddressContains($email, 'to', 'vinny@fandf.com');
```
coldic3: Can it be done that way?
```suggestion
        if (!self::isSymfonyMailerAvailable()) {
            /** @var EmailCheckerInterface $emailChecker */
            $emailChecker = $this->get('sylius.behat.email_checker');
            Assert::assertTrue($emailChecker->hasRecipient('vinny@fandf.com'));
        } else {
            $email = self::getMailerMessage();
            self::assertEmailAddressContains($email, 'to', 'vinny@fandf.com');
        }
```
Thanks to this we don't need to skip the test.

tests/Controller/Utils/MailerAssertionsTrait.php
```diff
@@ -0,0 +1,98 @@
...
+    {
+        return self::getMessageMailerEvents()->getMessages($transport);
+    }
+
+    public static function getMailerMessage(int $index = 0, string $transport = null): ?RawMessage
+    {
+        return self::getMailerMessages($transport)[$index] ?? null;
+    }
+
+    private static function getMessageMailerEvents(): MessageEvents
```
GSadee, line 64: private methods should be after the public ones

GSadee (approved):

GSadee:
Thank you, Kamil! :tada:

PR #735, opened 2022-12-14
Drop Symfony 4.4 support

TheMilek:
We will no more support Symfony 4.4

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,9 +19,9 @@ jobs:
             fail-fast: false
             matrix:
                 php: [8.0, 8.1]
-                node: [10.x]
+                node: [16.x]
                 mysql: [5.7, 8.0]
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -19,9 +19,9 @@ jobs:
             fail-fast: false
             matrix:
                 php: [8.0, 8.1]
-                node: [10.x]
+                node: [16.x]
                 mysql: [5.7, 8.0]
-                symfony: [^4.4, ~5.2.0, ~5.3.0]
+                symfony: [^5.2]
```
GSadee, line 24: With this provision, we are only testing 5.4 anyway, but as those versions (Sf 5.2 and 5.3) have not been supported for at least a year anyway, I think we can remove support for them clearly, as we did in Sylius 1.11

GSadee (approved):

GSadee:
Thank you, Kamil! :1st_place_medal:

PR #745, opened 2025-02-10
Marking project as deprecated

mamazu:


README.md
```diff
@@ -1,18 +1,13 @@
...
-          <source media="(prefers-color-scheme: light)" srcset="https://media.sylius.com/sylius-logo-800.png">
-          <img alt="Sylius Logo." src="https://media.sylius.com/sylius-logo-800.png">
-        </picture>
-    </a>
-</p>
-
 <h1 align="center">Sylius Shop API </h1>
 
+> [!WARNING]  
+> This project is deprecated. The new way of building APIs in Sylius is with [ApiPlatform](https://docs.sylius.com/the-customization-guide/customizing-api).
```
diimpp: - Entry url for api should be https://docs.sylius.com/getting-started-with-sylius/using-api but it sounds good to reference customization guide as well
- I would rather call it `Sylius Api`, than `ApiPlatform Api`

```suggestion
> This plugin is deprecated. The new way of building APIs in Sylius is with [Sylius Api](https://docs.sylius.com/getting-started-with-sylius/using-api).
```

damonsson (approved):

